PR #1, opened 2022-04-18
Initial commit of Zephyr ported WPA Supplicant

sr1dh48r:


zephyr/Kconfig
```diff
@@ -7,6 +7,6 @@
 
 menuconfig WPA_SUPP
         bool "WPA supplicant"
-	default y
+	default n
```
rlubos, line 10: `default n` is implicit, no need to specify it.

zephyr/west.yml
```diff
@@ -0,0 +1,16 @@
...
+  self:
+    path: wpa_supplicant_zephyr
+
+  remotes:
+    - name: zephyrproject-rtos
+      url-base: https://github.com/zephyrproject-rtos
+
+  projects:
+    - name: zephyr
+      remote: zephyrproject-rtos
```
rlubos, line 14: This is a bit confusing, why is the module pulling the upstream Zephyr? It should rather work with `sdk-zephyr` pulled by `sdk-nrf` manifest file.

Kconfig.sdk-hostap
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2019 - 2020 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menu "WPA supplicant"
+
+rsource "zephyr/Kconfig"
```
rlubos, line 9: I wonder why are there two separate Kconfig files, can't we just keep the configuration in a single file?
I also don't see `zephyt/module.yml` here, which specifies where's module Kconfig and CMakeLists.txt file (see nrfxlib for example: https://github.com/nrfconnect/sdk-nrfxlib/blob/main/zephyr/module.yml)

src/l2_packet/l2_packet_zephyr.c
```diff
@@ -0,0 +1,225 @@
+/*
+ * WPA Supplicant - Layer2 packet handling for ZephyrOS
+ * Copyright (c) 2022 Jouni Malinen <j@w1.fi>
```
rlubos, line 3: Is the copyright correct here?

zephyr/prj.conf
```diff
@@ -0,0 +1,58 @@
+CONFIG_EARLY_CONSOLE=y
```
rlubos, line 1: It's a bit confusing to me, why is there `prj.conf` inside a module? Any sample demonstrating supplicant should be part of `sdk-nrf`.

zephyr/prj_qemu.conf
```diff
@@ -0,0 +1 @@
+CONFIG_TEST_RANDOM_GENERATOR=y
```
rlubos, line 1: If some specific configuration needs to be applied for qemu, you can add `boards/qemu_x86.conf` file to the sample, it'll be applied automatically for that platform.

zephyr/README.rst
```diff
@@ -0,0 +1,33 @@
+.. _hello_world:
```
rlubos, line 1: Copy-paste

zephyr/src/supp_mgmt.h
```diff
@@ -0,0 +1,106 @@
+/*
```
rlubos, line 1: So, is this really different from Zephyr's wifi mgmt layer?

rlubos (commented):
Some initial, generic feedback

zephyr/src/supp_main.c
```diff
@@ -0,0 +1,196 @@
+/*
```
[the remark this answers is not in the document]
wbober: Ah my bad. You are right. The new UI mislead me and I didn't spot it's `zephyr/src`...

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+set(WPA_SUPPLICANT_SRC ${CMAKE_CURRENT_SOURCE_DIR}/../wpa_supplicant)
+set(HOSTAP_BASE ${CMAKE_CURRENT_SOURCE_DIR}/../)
+
+set(CMAKE_EXE_LINKER_FLAGS "--specs=nosys.specs -lnosys")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMISSING_SYSCALL_NAMES")
+
```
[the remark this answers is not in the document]
sr1dh48r: done.

zephyr/src/supp_main.c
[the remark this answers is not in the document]
sr1dh48r: Assuming nothing to be done here.

zephyr/west.yml
sr1dh48r: Removing leftover files.

Kconfig.sdk-hostap
sr1dh48r: Updated to include module.yml and single KConfig.

src/l2_packet/l2_packet_zephyr.c
sr1dh48r: Updated.

zephyr/prj.conf
sr1dh48r: File removed.

zephyr/README.rst
sr1dh48r: File removed.

zephyr/src/supp_mgmt.h
sr1dh48r: It's a copy of the wifi_mgmt.h but uses a different _NET_SUPP_CODE to avoid conflict with the wifi_mgmt enums. We are going update this to include the full WPA_CLI commands that supplicant supports going forward.

zephyr/prj_qemu.conf
sr1dh48r: Removed the file

src/drivers/driver_zephyr.c
```diff
@@ -0,0 +1,540 @@
...
+K_MBOX_DEFINE(wpa_supp_mbox);
+
+static void wpa_supp_drv_mbox_msg_handler(void *eloop_ctx, void *timeout_ctx)
+{
+	struct k_mbox_msg msg;
+	struct zep_wpa_supp_mbox_msg_data mbox_msg_data;
+
+	/* Prepare to receive message */
+	msg.info = 10000;
+	msg.size = 10000;
```
rlubos, line 22: Why are those hardcoded to such value? Should this rather reflect the expected message size?

zephyr/src/supp_mgmt.h
rlubos: I still strongly encourage to try to extend the actual wifi mgmt layer from Zephyr when we have a clear image on what is needed.

rlubos (commented):
There's still quite a few TODOs left, I assume they're not going to be addressed for the initial delivery?

wbober:
@sr1dh48r How do you actually test this? Are we missing a basic wifi sample for this?

zephyr/src/supp_mgmt.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+#ifndef __SUPP_MGMT_H__
+#define __SUPP_MGMT_H__
+
+/* Management part definitions */
+#include <net/net_mgmt.h>
+#include <net/wifi.h>
+
+#define _NET_SUPP_LAYER NET_MGMT_LAYER_L2
```
rlubos, line 14: After some consideration with @wbober we're starting to wonder if there's even a point introducing a supplicant-specific net mgmt layer in NCS.

The core purpose of wifi_mgmt layer was to abstract the MGMT API , so that it could be implemented for various devices. That way an application written with WiFi in mind could use this abstraction to implement an app portable to different devices.

Here however, we're introducing another abstraction layer, that will work only with supplicant and nRF devices. That doesn't make much sense on the second thought.

Perhaps, instead of increasing the complexity with yet another abstraction layer, the application could call the `wpa_supplicant` directly? That way, we would not put on ourselves any constraints applied by the wifi_mgmt API, and we'd avoid extra complexity of introducing yet another network management layer.

In the future, we might even consider implementing a minimalistic wifi_mgmt layer, based on the upstream APIs, on top of the `wpa_supplicant` API. Yes, the API might have less capabilities then we expect from it (unless we want to extend it), but at least we'd retain compability with the upstream at low cost.

src/drivers/driver_zephyr.c
```diff
@@ -0,0 +1,540 @@
...
+
+	if_ctx = os_zalloc(sizeof(*if_ctx));
+
+	if (if_ctx == NULL) {
+		return NULL;
+	}
+
+	if_ctx->supp_if_ctx = ctx;
+
+	if_ctx->dev_ctx = net_if_get_device(net_if_get_default());
```
wbober, line 210: @rlubos Is this appropriate? What if there are multiple interfaces, e.g., Thread and WiFi? Will this still work?

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+	CONFIG_NO_STDOUT_DEBUG
+	)
+
+zephyr_library_include_directories(
+    	${HOSTAP_BASE}/
+	${WPA_SUPPLICANT_SRC}/
+	${HOSTAP_BASE}/src/utils
+	${HOSTAP_BASE}/src/drivers
+	${HOSTAP_BASE}/src
+	${GNUARMEMB_TOOLCHAIN_PATH}/opt/arm-none-eabi/include
```
wbober, line 43: @rlubos Do you think this is needed here?

src/drivers/driver_zephyr.c
rlubos: Good point, this won't be reliable in case of multiple network interfaces. We should likely use `device_get_binding()` here, in Zephyr interface names corresponds to the network device name.

zephyr/CMakeLists.txt
rlubos: I don't think it should be needed, it should already be provided in the include path by Zephyr.

sr1dh48r:
> @sr1dh48r How do you actually test this? Are we missing a basic wifi sample for this?

@wbober We have pushed the sample application to nrf/samples/wifi/shell

src/crypto/tls_mbedtls.c
```diff
@@ -0,0 +1,1181 @@
...
+		with = "";
+	len_with = strlen(with);
+
+	// count the number of replacements needed
+	ins = orig;
+	for (count = 0; (tmp = strstr(ins, rep)); ++count) {
+		ins = tmp + len_rep;
+	}
+
+	tmp = result = malloc(strlen(orig) + (len_with - len_rep) * count + 1);
```
rlubos, line 913: Shouldn't this use `os_zalloc()`?

src/crypto/tls_mbedtls.c
```diff
@@ -0,0 +1,1181 @@
...
+	}
+
+	os_strlcpy(buf, name, buflen);
+
+	// Translate to common format for hwsim tests to pass
+	strremove(buf, "TLS-");
+	strremove(buf, "WITH-");
+	char *tmp = str_replace(buf, "AES-", "AES");
+	os_memcpy(buf, tmp, buflen);
+	free(tmp);
```
rlubos, line 954: Ditto, `os_free()`?

zephyr/src/supp_mgmt.c
```diff
@@ -0,0 +1,380 @@
...
+
+	}
+
+	wpa_supplicant_enable_network(wpa_s_0, ssid);
+	wpa_supplicant_select_network(wpa_s_0, ssid);
+
+	return 0;
+}
+
+NET_MGMT_REGISTER_REQUEST_HANDLER(NET_REQUEST_SUPP_SCAN, wifi_supp_scan);
```
rlubos, line 363: Weren't we supposed to drop the net mgmt use?

zephyr/Kconfig
```diff
@@ -0,0 +1,44 @@
...
+# Copyright (c) 2022 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config WPA_SUPP
+    bool "WPA WPA_SUPP"
+    select EARLY_CONSOLE
+    select NET_SOCKETS
+    select NET_SOCKETS_POSIX_NAMES
```
markaj-nordic, line 12: This setting may raise conflicts in case the application is built with POSIX support (CONFIG_POSIX_API). Any reason for not using POSIX in supplicant?

rlubos: You could use `zsock_*` namespace for socket operations to avoid posix dependencies, that's what some Zephyr libraries do.

zephyr/Kconfig
```diff
@@ -0,0 +1,44 @@
+# Nordic WiFi driver for nRF52840 and nRF5340
+#
+# Copyright (c) 2022 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config WPA_SUPP
+    bool "WPA WPA_SUPP"
+    select EARLY_CONSOLE
```
rlubos, line 10: Why does supplicant select this config, is it really essential for supplicant operation?

zephyr/Kconfig
```diff
@@ -0,0 +1,44 @@
...
+    select NET_SOCKETS
+    select NET_SOCKETS_POSIX_NAMES
+    select NET_SOCKETS_PACKET
+    select NET_MGMT_EVENT_INFO
```
rlubos, line 14: If you select `NET_MGMT_EVENT_INFO`, you should also select its dependencies, i. e. `NET_MGMT` and `NET_MGM_EVENT`.

zephyr/Kconfig
markaj-nordic: This could be an option in case the dependencies are located only in Zephyr platform specific code but I got an impression that some generic supplicant modules are affected as well (not 100% sure though). Another option as proposed by @rlubos  is letting the application to chose API by having:

`depends on NET_SOCKETS_POSIX_NAMES || POSIX_API`

but then the supplicant would need to support both cases. The general thought is that the supplicant should not forcefully imply POSIX_NAMES nor POSIX_API.

krish2718: In general agree with the suggestion, but the reason for forcing `NET_SOCKETS_POSIX_NAMES` is to avoid changing upstream supplicant code, as it doesn't have os_* wrappers for socket APIs. So, unless we fix that first and upsteam, we need to live with this.

Damian-Nordic: Ok, but `POSIX_API` serves a similar purpose as `NET_SOCKETS_POSIX_NAMES` and the former brings a bit more POSIX features, so we can support both flags by removing the `select` and applying the following patch:
```
--- modules/lib/hostap/src/utils/includes.h
+++ modules/lib/hostap/src/utils/includes.h
@@ -44,9 +44,16 @@
 #endif /* CONFIG_NATIVE_WINDOWS */
 
 #if defined(CONFIG_ZEPHYR)
+#if defined(CONFIG_POSIX_API)
+#include <posix/arpa/inet.h>
+#include <posix/sys/select.h>
+#include <posix/sys/socket.h>
+#include <posix/unistd.h>
+#else /* defined(CONFIG_POSIX_API) */
 #include <net/net_ip.h>
 #include <net/socket.h>
+#endif /* defined(CONFIG_POSIX_API) */
 #include <shell/shell.h>
-#endif
+#endif /* defined(CONFIG_ZEPHYR) */
```

We can also add something like:
```
#if !defined(CONFIG_POSIX_API) && !defined(CONFIG_NET_SOCKETS_POSIX_NAMES)
#pragma error "You must enable either ..."
#endif
```
to better inform a user what is missing in their config.

We use `POSIX_API` in Matter project for exactly the same reason - be able to use platform-independent code shared across Zephyr, Linux, Darwin etc, so it's kinda crucial that WPA supplicant build with that flag :)

krish2718: Thanks for the patch. I tried this, but its unable to find `open` definition, so, build fails. `posix/unistd.h` and `lib/os/fdtable.c` also doesn't have a definition for `open`. Am I missing something?

```
libc.a(lib_a-openr.o): in function `_open_r':
openr.c:(.text._open_r+0x14): undefined reference to `_open'
```

src/utils/os_zephyr.c
```diff
@@ -0,0 +1,279 @@
...
+#endif
+	struct timespec ts;
+	int res;
+
+	if (TEST_FAIL()) {
+		return -1;
+	}
+
+	while (1) {
+		res = clock_gettime(clock_id, &ts);
```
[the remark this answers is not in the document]
krish2718: This was taken from `os_unix.c`, the loop is just to try fallback mechanism for various clocks (BOOTTIME, MONOTONIC and REALTIME), the worst case scenario is if REALTIME fails, in which case the loop exits.

zephyr/Kconfig
krish2718: Its not `fopen` but the `open` system call that is being used and as per https://docs.zephyrproject.org/latest/services/portability/posix.html it is supported.

> Shouldn't we set CONFIG_NO_VLAN for Zephyr or rewrite config_file.c?
Yes, we are taking care of the defined in CMakeLists.txt as they are static and not use configurable.

Damian-Nordic: Maybe the documentation is wrong. I can't find a declaration of `open` in Zephyr's `unistd.h` nor elsewhere. Also, I checked with `puncover` tool that the only caller of `open` syscall is `fopen`. Is the tool wrong and you call `open` directly somewhere in `hostap` module?

krish2718: Yes, even I was unable to find a definition. Most code that uses `fopen` in hostapd is unused (except for one call in `src/crypto/random.c`, we directly use `open` for open device files `src/crypto/random.c`, commenting that out using `CONFIG_NO_RANDOM_POOL` builds fine. and we can proceed with that for now, but just wanted to check `open` syscall status in Zephyr.

krish2718:
Closed by mistake.

zephyr/src/supp_mgmt.h
krish2718: As per offline discussions. This has now been removed, shell directly talks to wpa_supplicant code.

zephyr/src/main.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef __MAIN_H__
+#define __MAIN_H__
+extern void start_wpa_supplicant_thread(void);
```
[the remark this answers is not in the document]
krish2718: renamed to supp_main.h as its related to supp_main.c

src/drivers/driver_zephyr.c
krish2718: Fixed to use proper interface name

zephyr/src/supp_mgmt.c
krish2718: dropped

zephyr/Kconfig
krish2718: Implemented support for both, and sample uses POSIX as default.

zephyr/Kconfig
krish2718: dropped

src/utils/os_zephyr.c
```diff
@@ -0,0 +1,279 @@
...
+
+	if (r && src) {
+		os_memcpy(r, src, len);
+	}
+	return r;
+}
+
+void *os_zalloc(size_t size)
+{
+	return calloc(1, size);
```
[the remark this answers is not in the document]
krish2718: calloc internally uses malloc, so, IIUC as per comment this is fine, hence closing it. We will review the dedicated heap for driver/supplicant needs in the future.

krish2718:
@rlubos @Damian-Nordic @wbober not sure about the process for resolving comments (nothing is mentioned at https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-guidelines), so, resolved all comments as corresponding code changes are done, please do a final review and feel free to re-open if you are not satisfied with the resolution.

Also, as requested (offline) squashed all commits into one and added all authors DCO. IMHO, this is good for an initial commit, and we will address few remaining involved changes in the near future.

zephyr/src/supp_wrapper.c
```diff
@@ -0,0 +1,583 @@
...
+#include "wifi_mgmt.h"
+
+//#define MAX_SSID_LEN 32
+
+int cli_main (int, const char **);
+extern struct wpa_supplicant *wpa_s_0;
+
+int zephyr_supp_disable_network(int id)
+{
+    char cmd[512] = {'\0'};
```
rlubos, line 36: Does this buffer really need to be that large? A lot of this space appears to be unused at the moment. Note that this is allocated on a stack so will automatically affect any thread that'd make use of those commands.

zephyr/src/supp_wrapper.c
```diff
@@ -0,0 +1,583 @@
...
+    char quote;
+    int ret = -1;
+
+    os_snprintf(cmd, sizeof(cmd), "disable_network %d", id);
+    argv[0] = cmd;
+    argvp = &argv[0];
+
+    quote = z_shell_make_argv(&argc, argvp, cmd, 4);
+
+	if (argc == 0) {
```
rlubos, line 49: You should decide whether you use tabs or spaces, a mixed up approach looks messy (at least in my environment, where tab == 8 spaces).

zephyr/src/supp_wrapper.c
```diff
@@ -0,0 +1,583 @@
...
+		return -ENOEXEC;
+	} else if ((argc == 1) && (quote != 0)) {
+		wpa_printf(MSG_ERROR, "not terminated: %c\n", quote);
+		return -ENOEXEC;
+	}
+
+    ret = cli_main (argc, argv);
```
rlubos, line 56: The space after `cli_main ` don't seem to be intended? There are other places like this as well.

zephyr/src/supp_wrapper.c
```diff
@@ -0,0 +1,583 @@
...
+    quote = z_shell_make_argv(&argc, argvp, cmd, 4);
+
+	if (argc == 0) {
+		return -ENOEXEC;
+	} else if ((argc == 1) && (quote != 0)) {
+		wpa_printf(MSG_ERROR, "not terminated: %c\n", quote);
+		return -ENOEXEC;
+	}
+	ret = cli_main (argc, argv);
+	return ret;
```
rlubos, line 112: So this piece of code seems to be duplicated across various functions, I wonder if it could be extracted into som helper function?

zephyr/src/supp_mgmt.h
rlubos: The header should also be removed then?

zephyr/Kconfig
```diff
@@ -0,0 +1,45 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config WPA_SUPP
+    bool "WPA WPA_SUPP"
+    depends on  NET_SOCKETS
+    depends on  NET_SOCKETS_POSIX_NAMES || POSIX_API
+    select NET_SOCKETS_PACKET
+    select NET_MGMT
```
rlubos, line 13: Is NET_MGMT dependency still required given that supp_mgmt has been dropped?

wpa_supplicant/wpa_supplicant.c
```diff
@@ -905,6 +905,22 @@ void wpa_supplicant_reinit_autoscan(struct wpa_supplicant *wpa_s)
 	}
 }
 
+static void dhcp_war(struct wpa_supplicant *wpa_s)
```
rlubos, line 908: For an outsider it might not be clear what is this workaround for, would be nice to add comment about the problem.

rlubos (commented):
What about all those TODO or FIXME comments, are they targetting the initial release?

zephyr/src/supp_shell.c
```diff
@@ -0,0 +1,420 @@
...
+}
+#endif /* notyet */
+
+static void supplicant_mgmt_event_handler(struct net_mgmt_event_callback *cb,
+					  uint32_t mgmt_event,
+					  struct net_if *iface)
+{
+	switch (mgmt_event) {
+#ifdef notyet
+	case NET_EVENT_SUPP_SCAN_RESULT:
```
rlubos, line 105: This could use some cleanup as well (I assume it's just some obsolete code?)

src/common/wpa_ctrl.c
```diff
@@ -35,7 +35,7 @@
 #include "wpa_ctrl.h"
 #include "common.h"
 
-
+#define inet_ntoa(a) ""
```
wbober, line 38: How about using net_addr_ntop?

wbober (approved):

zephyr/src/supp_wrapper.c
```diff
@@ -0,0 +1,583 @@
...
+#include "p2p_supplicant.h"
+#include "wpa_supplicant_i.h"
+#include "wifi_mgmt.h"
+
+//#define MAX_SSID_LEN 32
+
+int cli_main (int, const char **);
+extern struct wpa_supplicant *wpa_s_0;
+
+int zephyr_supp_disable_network(int id)
```
wbober, line 34: @krish2718 From the application developer perspective this is the public API for supplicant, right? If so aren't we then missing an include file?

krish2718: This is still WIP, as you can see this is not called anywhere, so, will remove this for now add it back once its complete.

wbober: @krish2718 I see now that I wasn't clear in my question. What I really ment was most functions in this file rather than this particular one.

zephyr/src/supp_wrapper.c
krish2718: This file was WIP, so, removed now, will be readded later.

zephyr/src/supp_main.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef __MAIN_H__
+#define __MAIN_H__
+extern void start_wpa_supplicant_thread(void);
```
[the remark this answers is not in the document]
krish2718: This was a relic when we used to manually invoke the wpa_supplicant, removed this now.

zephyr/src/supp_wrapper.c
krish2718: Got it, this file is now removed, and will be re-worked later.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,326 @@
...
+	CONFIG_NO_CONFIG_BLOBS
+	CONFIG_WPA_S_ZEPHYR_L2_WIFI_MGMT
+	#CONFIG_NO_STDOUT_DEBUG
+	CONFIG_CTRL_IFACE
+	CONFIG_CTRL_IFACE_UDP
+	CONFIG_NO_RANDOM_POOL
+	)
+
+zephyr_library_include_directories(
+    ${HOSTAP_BASE}/
```
markaj-nordic, line 42: Duplicated line?

wpa_supplicant/wpa_cli.c
```diff
@@ -217,31 +227,36 @@ static void wpa_cli_close_connection(void)
...
 	size_t len;
 	int ret;
 
+    buf = os_zalloc(sizeof(u8) * 4096);
+	if (!buf){
+		printf ("Failed to allocate mem for command buf\n");
+		return -1;
+	}
 	if (ctrl_conn == NULL) {
 		printf("Not connected to wpa_supplicant - command dropped.\n");
```
markaj-nordic, line 250: Looks like leakage, shouldn't we free the `buf` here? And even if this condition is not hit, where this buffer is going to be freed?

wpa_supplicant/wpa_cli.c
```diff
@@ -217,31 +227,36 @@ static void wpa_cli_close_connection(void)
...
-
+#ifdef INTERACTIVE
 static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print)
 {
-	char buf[4096];
+	char * buf = NULL;
 	size_t len;
 	int ret;
 
+    buf = os_zalloc(sizeof(u8) * 4096);
```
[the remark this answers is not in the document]
markaj-nordic: Isn't it already defined as `CMD_BUF_LEN`?

wpa_supplicant/wpa_supplicant.c
```diff
@@ -905,6 +905,27 @@ void wpa_supplicant_reinit_autoscan(struct wpa_supplicant *wpa_s)
 	}
 }
 
+// TODO: This WAR is needed as we always lose the first frame after association (DHCP),
```
markaj-nordic, line 908: Is this a final workaround or there is some kind of different remedy pending for that?

Damian-Nordic (approved):

wpa_supplicant/wpa_supplicant.c
krish2718: Yes, there is a bug against sheliak, but as its a ROM code, will take time, so, this is the final WAR for now.

wpa_supplicant/wpa_supplicant.c
```diff
@@ -6806,7 +6828,7 @@ static int wpa_supplicant_init_iface(struct wpa_supplicant *wpa_s,
 		return -1;
 	wpa_sm_set_eapol(wpa_s->wpa, wpa_s->eapol);
 
-	wpa_s->ctrl_iface = wpa_supplicant_ctrl_iface_init(wpa_s);
+	wpa_s->ctrl_iface = (void *)-1 ;//wpa_supplicant_ctrl_iface_init(wpa_s);
```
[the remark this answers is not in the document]
krish2718: Used proper define

markaj-nordic (approved):

rlubos (approved):

sachinthegreen (approved):

PR #9, opened 2022-07-20
Fix l2 packet

krish2718:


rlubos (approved):

sachinthegreen (approved):

rlubos:
When, if you're adding fixes to `sdk-hostap` you should also open a PR in `sdk-nrf` to update `sdk-hostap` revision - otherwise the fixes are not avaiable to NCS users.

krish2718:
> When, if you're adding fixes to `sdk-hostap` you should also open a PR in `sdk-nrf` to update `sdk-hostap` revision - otherwise the fixes are not avaiable to NCS users.

Thanks, already doing that.

PR #11, opened 2022-07-29
drivers: zephyr: use struct device instead of name

gmarull:
Use device references instead of interface name, so that we can avoid
usage of device_get_binding.

### DIFF
--- a/src/drivers/driver_zephyr.h
+++ b/src/drivers/driver_zephyr.h
@@ -69,7 +69,7 @@ struct zep_wpa_supp_dev_callbk_fns {
 
 struct zep_wpa_supp_dev_ops {
 	void *(*init)(void *supp_drv_if_ctx,
-		      const char *iface_name,
+		      const struct device *dev,
 		      struct zep_wpa_supp_dev_callbk_fns *callbk_fns);
 	void (*deinit)(void *if_priv);
 	int (*scan2)(void *if_priv,


rlubos (commented):
I think those changes are not enough, this init function is called here if I undestand correctly:
https://github.com/nrfconnect/sdk-hostap/blob/main/src/drivers/driver_zephyr.c#L294
And AFAIK `wpa_drv_zep_init()` is an API defined by the supplicant itself, so I don't think there's much to do about it.

gmarull:
> I think those changes are not enough, this init function is called here if I undestand correctly: [https://github.com/nrfconnect/sdk-hostap/blob/main/src/drivers/driver_zephyr.c#L294](https://github.com/nrfconnect/sdk-hostap/blob/main/src/drivers/driver_zephyr.c?rgh-link-date=2022-07-29T09%3A43%3A53Z#L294) And AFAIK `wpa_drv_zep_init()` is an API defined by the supplicant itself, so I don't think there's much to do about it.

I see, thanks for the details. However, I imagine there must be a way to stay away from device_get_binding somehow.

krish2718:
> > I think those changes are not enough, this init function is called here if I undestand correctly: [https://github.com/nrfconnect/sdk-hostap/blob/main/src/drivers/driver_zephyr.c#L294](https://github.com/nrfconnect/sdk-hostap/blob/main/src/drivers/driver_zephyr.c?rgh-link-date=2022-07-29T09%3A43%3A53Z#L294) And AFAIK `wpa_drv_zep_init()` is an API defined by the supplicant itself, so I don't think there's much to do about it.
> 
> I see, thanks for the details. However, I imagine there must be a way to stay away from device_get_binding somehow.

In Wi-Fi including Wpa_supplicant, interface name is used as input, so, if we don't want to use that we need to replace that with some other identifier esp. when we add multi-interface support.

sachinthegreen (approved):

gmarull:
> @gmarull as per your commebt above can we close this?

yes, let's close this for now. Note that device labels have been deprecated upstream, so device_get_binding will only accept manual names from `DEVICE_DEFINE` or DT node names, e.g. `gpioa@abcd1234`.

PR #12, opened 2022-08-04
Compile out logs based on configured log level

Damian-Nordic:
1. Add WPA_SUPP_LOG_LEVEL_* Kconfig options to control the WPA supplicant log level the same way as other SDK components.
2. Add WPA_SUPP_DEBUG_LEVEL Kconfig option to be used for compile-time filtering of WPA supplicant debug messages. By default, it is aligned with WPA_SUPP_LOG_LEVEL.
3. Implement Zephyr variants of wpa_debug.h and wpa_debug.c files that use Zephyr logging subystem as the default output and apply compile-time filtering for the messages.

This reduces the wifi shell sample size from 626884 to 543196 bytes when `WPA_SUPP_LOG_LEVEL_ERR` is enabled.

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>

### DIFF
--- a/src/utils/common.h
+++ b/src/utils/common.h
@@ -532,7 +532,7 @@ static inline int is_multicast_ether_addr(const u8 *a)
 
 #define broadcast_ether_addr (const u8 *) "\xff\xff\xff\xff\xff\xff"
 
-#include "wpa_debug.h"
+#include <utils/wpa_debug.h>
 
 
 struct wpa_freq_range_list {
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -20,7 +20,7 @@ zephyr_compile_definitions(
 )
 
 zephyr_include_directories(
-	${CMAKE_CURRENT_SOURCE_DIR}/
+	${CMAKE_CURRENT_SOURCE_DIR}/src
 	${HOSTAP_BASE}/
 	${WPA_SUPPLICANT_BASE}/
 	${COMMON_SRC_BASE}/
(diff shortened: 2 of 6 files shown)


krish2718:
Current policy it to keep the changes to core hostap code minimal, so, any customizations are handled either using CONFIG_ZEPHYR or having a dedicated file either in hostap/zephyr folder. 

So, can we duplicate the wpa_debug.c/h to zephyr directory and adjust CMakelists.txt to pick up from there? This gives us freedom to whatever we want and all our changes are neatly in the zephyr directory.

krish2718 (changes requested):
Thanks for picking this up.

Current policy is to keep the changes to the core `hostap` code minimal, and any customisations are handled either using CONFIG_ZEPHYR in `hostap` code or having a dedicated file either in `hostap/zephyr` folder.

So, can we duplicate the `wpa_debug.c/h` to zephyr directory and adjust `CMakelists.txt` to pick up from there? This gives us freedom to do whatever we want and all our changes are neatly wrapped in the zephyr directory.

Damian-Nordic:
@krish2718 yes, makes sense. I'll re-do it in that fashion.

src/utils/common.h
```diff
@@ -532,7 +532,7 @@ static inline int is_multicast_ether_addr(const u8 *a)
 
 #define broadcast_ether_addr (const u8 *) "\xff\xff\xff\xff\xff\xff"
 
-#include "wpa_debug.h"
+#include <utils/wpa_debug.h>
```
krish2718, line 535: if you move from zephyr/src/utils -> zephyr/src then I guess this isn't needed?

Damian-Nordic: It is needed. Unfortunately, using `#include "file.h"` type of includes always prefers a local file, so I would still need to change it to `<wpa_debug.h>`.

krish2718: ok, thanks.

krish2718 (approved):
In my build with default values it saves about 62K.

sachinthegreen (approved):

wbober (approved):

LuDuda (approved):
LGTM :+1:

rlubos (approved):

PR #14, opened 2022-08-23
Merge dev branch

krish2718:
This is in preparation for NCS 2.1.0 release, as DEV has diverged the commits are high, but in future where we move to feature branches from DEV branch as rest of Nordic, future PRs are smaller.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -110,6 +110,7 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
 				     bool more_res)
 {
 	struct wpa_scan_res **tmp = NULL;
+	size_t scan_res_len  = sizeof(struct wpa_scan_res) + r->ie_len + r->beacon_ie_len;
 
 	tmp = os_realloc_array(if_ctx->scan_res2->res,
 			       if_ctx->scan_res2->num + 1,
@@ -121,7 +122,15 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
 		goto err;
 	}
 
-	tmp[if_ctx->scan_res2->num++] = r;
+	struct wpa_scan_res *sr = os_zalloc(scan_res_len);
+	if (!sr) {
+		wpa_printf(MSG_ERROR, "%s: Failed to alloc scan results(%d bytes)\n", __func__, scan_res_len);
+		goto err;
+	}
+
+	os_memcpy(sr, r, scan_res_len);
+
+	tmp[if_ctx->scan_res2->num++] = sr;
 
 	if_ctx->scan_res2->res = tmp;
 
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -112,6 +112,9 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
 	struct wpa_scan_res **tmp = NULL;
 	size_t scan_res_len  = sizeof(struct wpa_scan_res) + r->ie_len + r->beacon_ie_len;
 
+	if (!if_ctx->scan_res2)
+		return;
+
 	tmp = os_realloc_array(if_ctx->scan_res2->res,
 			       if_ctx->scan_res2->num + 1,
 			       sizeof(struct wpa_scan_res *));
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -484,6 +484,7 @@ static int wpa_drv_zep_authenticate(void *priv,
 {
 	struct zep_drv_if_ctx *if_ctx = NULL;
 	const struct zep_wpa_supp_dev_ops *dev_ops = NULL;
+	struct wpa_bss *curr_bss;
 	int ret = -1;
 
 	if ((!priv) || (!params)) {
@@ -501,8 +502,17 @@ static int wpa_drv_zep_authenticate(void *priv,
 	
 	if_ctx->ssid_len = params->ssid_len;
 
+	curr_bss = wpa_bss_get(if_ctx->supp_if_ctx, params->bssid, params->ssid, params->ssid_len);
+
+	if (!curr_bss) {
+		wpa_printf(MSG_ERROR, "%s: Failed to get BSS", __func__);
+		ret = -1;
+		goto out;
+	}
+
 	ret = dev_ops->authenticate(if_ctx->dev_priv,
-				    params);
+			    params,
+			    curr_bss);
 
 	if (ret) {
 		wpa_printf(MSG_ERROR, "%s: authenticate op failed\n", __func__);
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -5,63 +5,34 @@
  * This software may be distributed under the terms of the BSD license.
  * See README for more details.
  */
+#include <zephyr/kernel.h>
+
 #include "includes.h"
 #include "utils/common.h"
 #include "eloop.h"
 #include "driver_zephyr.h"
+#include "supp_main.h"
 
-K_MBOX_DEFINE(wpa_supp_mbox);
 #define SCAN_TIMEOUT 30
 
-static void wpa_supp_drv_mbox_msg_handler(void *eloop_ctx,
-					  void *timeout_ctx)
+void wpa_supplicant_event_wrapper(void *ctx,
+				enum wpa_event_type event,
+				union wpa_event_data *data)
 {
-	struct k_mbox_msg msg;
-	struct zep_wpa_supp_mbox_msg_data mbox_msg_data;
+	struct wpa_supplicant_event_msg msg = { 0 };
 
-	/* Prepare to receive message */
-	msg.size = sizeof(mbox_msg_data);
-	msg.rx_source_thread = K_ANY;
-
-	k_mbox_get(&wpa_supp_mbox,
-		   &msg,
-		   &mbox_msg_data,
-		   K_FOREVER);
-
-	mbox_msg_data.cb(mbox_msg_data.ctx,
-			 mbox_msg_data.data,
-			 0);
+	msg.ctx = ctx;
+	msg.event = event;
+	if (data) {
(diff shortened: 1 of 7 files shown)


zephyr/src/supp_main.c
```diff
@@ -157,11 +252,12 @@ static void start_wpa_supplicant(void)
...
 			break;
 		}
-		wpa_s_0 = wpa_s;
-		wpa_s_0->conf->filter_ssids = 1;
-		wpa_s_0->conf->ap_scan= 1;
+		wpa_s->conf->filter_ssids = 1;
+		wpa_s->conf->ap_scan= 1;
 	}
 
+	register_dummy_socket();
```
wbober, line 259: I would like to better understand what proble the dummy socket addition is attempting to solve.

src/drivers/driver_zephyr.c
```diff
@@ -5,64 +5,99 @@
...
-}
+		if (mbox_msg_data.data)
+			os_free(mbox_msg_data.data);
+	}
 
+	return 0;
+}
 
-void wpa_supp_event_handler(void *ctx,
+K_THREAD_DEFINE(drv_events_thread,
```
wbober, line 59: I would like to better understand the issue you are trying to solve by adding this thread. I have some suspicion and also concern that this might not be the right way to go.

I think the key to properly passing events from the driver to supplicant is in `epool.c` which, at the moment, isn't thread safe.

I think it's going to be easier to talk offline on teams. Please reach me.

wbober (changes requested):

zephyr/src/zephyr_supp_api.c
```diff
@@ -0,0 +1,748 @@
...
+#include "fst/fst.h"
+#include "p2p_supplicant.h"
+#include "wpa_supplicant_i.h"
+#include "wifi_mgmt.h"
+
+int cli_main(int, const char **);
+extern struct wpa_supplicant *wpa_s_0;
+
+
+int zep_supp_ctrl_iface_status(struct wpa_supplicant *wpa_s,
```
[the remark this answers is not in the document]
krish2718: Most of code in this file is removed in this https://github.com/krish2718/sdk-hostap/blob/nrf_wifi_mgmt/zephyr/src/zephyr_supp_api.c the reason for merging this soon to be removed code is to get the driver changes from DEV branch in, and then immediately when the API changes in Zephyr are merged, the corresponding API changes from https://github.com/krish2718/sdk-hostap/blob/nrf_wifi_mgmt/zephyr/src/zephyr_supp_api.c will be merged and should address most comments.

zephyr/src/zephyr_supp_api.c
```diff
@@ -0,0 +1,748 @@
...
+		wpa_printf(MSG_ERROR, "not terminated: %c\n", quote);
+		return -ENOEXEC;
+	}
+
+	return cli_main(argc, argv);
+}
+
+int zephyr_supp_disable_network(int id)
+{
+	char cmd[512] = {'\0'};
```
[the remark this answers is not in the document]
krish2718: https://github.com/nrfconnect/sdk-hostap/pull/14#discussion_r955151831

src/drivers/driver_zephyr.c
```diff
@@ -5,64 +5,99 @@
...
 	struct zep_wpa_supp_mbox_msg_data mbox_msg_data;
 
 	mbox_msg_data.ctx = ctx;
-	mbox_msg_data.data = data;
-	mbox_msg_data.cb = cb;
+	mbox_msg_data.event = event;
+	mbox_msg_data.data = NULL;
+
+	if (data) {
+		mbox_msg_data.data = os_zalloc(data_len);
```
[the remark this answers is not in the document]
krish2718: Sure, we are just follow the wpa_supplicant notation, but in this file we can directly use Zephyr calls.

src/drivers/driver_zephyr.c
krish2718: Ideally we should follow Linux style and use sockets to communicate with driver, that avoid this extra thread and also the need for dummy socket. This is already planned in the roadmap.

zephyr/src/supp_main.c
krish2718: It's a bit complicated but let me take stab:

* Supplicant uses ELOOP mechanism to perform most tasks, and the default task is a 10s cleanup task
* So, if nothing else is active, `select/poll` is blocked for 10secs (the earliest time or one of the registered FDs is ready), in Linux this addressed as it uses sockets for communication (see below) and that will set the FD ready, so, no need to wait for 10s expiry.
    - netlink sockets for all kernel communication so whenever an event is received
    - uses sockets from user interface to core supplicant code

In Zephyr for above cases we use direct function calls, so, there is no event to unblock `select/poll`, so, for every task there is a delay, this causes the overall connection time to ~30secs, with the dummy socket making the FD ready after all tasks, this comes down to ~5secs.

Of course, this is just a WAR, we are working on a proper design that fits this wpa_supplicant design, which should be available in near future and then this WAR can be removed.

zephyr/src/zephyr_supp_api.c
```diff
@@ -0,0 +1,748 @@
+/**
+ * Copyright (c) 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/*
+ * WPA Supplicant / main() function for Zephyr OS
```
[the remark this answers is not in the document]
krish2718: oops, will fix.

src/drivers/driver_zephyr.c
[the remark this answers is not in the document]
krish2718: How hard are you on this? IMHO, wpa_supplicant in general uses OS wrapper even in OS specific code, e.g., see `driver_nl80211.c` this just makes a clean design and easier to re-use code.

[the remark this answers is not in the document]
wbober: Sure, that is fine.

I have a different view but it is subjective :)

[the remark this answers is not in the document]
krish2718: Thanks, lets park this for now then.

src/drivers/driver_zephyr.c
wbober: The problem I see is that you are manipulating non thread safe common structures from two different threads. My understanding goes like this:

* Supplicant eloop runs in `wpa_s_tid`.
* When an event from the driver comes it is put in the mbox. The `drv_events_thread` kicks-in and in turn runs `wpa_supplicant_event` in it's context. 
* The `wpa_supplicant_event` called from `drv_events_thread` manipulates supplicant internal state but it might happen that this is preempted by `wpa_s_tid` and leads to inconsistent state or crash.

Since you have already introduced a socket to trigger eloop processing why not to actually use it?

* Move the dummy socket logic to `driver_zephyr`.
* Remove the extra thread and mbox structure
* When there is an event to handle simply take `mbox_msg_data` and send it over the socket
* In the eloop socket handler read the data and pass it to `wpa_supplicant_event`. It is going to be processed in the correct context.

You probably don't need to set the socket nonblocking. Seems that if you set CONFIG_ELOOP_SELECT the eloop waits on the socket with a timeout.

krish2718: This is exactly what we had in the plan, extending dummy socket to real socket. In fact once we switch to sockets we don't even know an extra thread `drv_events_thread` (this was introduced to terminate the callstack).

> The wpa_supplicant_event called from drv_events_thread manipulates supplicant internal state but it might happen that this is preempted by wpa_s_tid and leads to inconsistent state or crash.

You are right, but we haven't seen any issue till now because, the `wpa_s_tid` thread blocks till `select` timesout, and by which time `wpa_supplciant_event` is long done, but there is a slight window where it can cause issues.

krish2718: Another reason why I haven't done full socket implementation for events right away is that, it has to work with multiple interfaces (probably with different drivers as well), in Linux this is achieved by netlink sockets, dummy sockets just uses a global UDP socket, so, we need to find a replacement for netlink in Zephyr (Per-interface Unix sockets would have been nice if we had Filesystem access).

wbober: Right. I think what I propose is a low hanging fruit; Then you could extend. 

Anyway. You are aware of the drawbacks and sdk-hostap is ultimately under your ownership, so I won't block.

wbober (approved):

src/drivers/driver_zephyr.c
krish2718: No, you are right, I have now implemented this and its working fine, it uses a global UDP socket but each message takes in the wpa_supplicant handle, so, it should work with multiple VIFs. We can work on converting UDP->Netlink replacement later.

krish2718: I have rebased as there are 4 commits in this area, so, please check 30159acf4 for this fix.

zephyr/src/supp_main.c
```diff
@@ -13,19 +13,31 @@
...
 #include "p2p_supplicant.h"
 #include "wpa_supplicant_i.h"
+#include "driver_i.h"
+
+#include "supp_main.h"
+
+
+#define DUMMY_SOCKET_PORT 9999
+
+#define DUMMY_SOCKET_PORT 9999
```
rlubos, line 36: This seems to be defined twice.

zephyr/src/supp_main.c
rlubos: Have you considered using a [socketpair](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/net/lib/sockets/socketpair.c) for this purpose, sounds like a better fit for this design than UDP socket (I am aware the documentation is not the best though, it seems that only a sample is documented). Just leaving this comment as FYI, I am aware we don't have time to change this before the release.

rlubos (approved):

zephyr/src/supp_main.c
krish2718: Urgh..fixed this in another branch, too many open branches :).

sachinthegreen (approved):

krish2718:
https://github.com/nrfconnect/sdk-nrf/pull/8495/checks?check_run_id=8090326640 CI passed

PR #15, opened 2022-08-27
Nrf wifi mgmt

krish2718:
This is based on #14 will be rebased on its merged.

### DIFF
--- a/zephyr/src/supp_api.c
+++ b/zephyr/src/supp_api.c
@@ -14,7 +14,6 @@
 #include "wpa_supplicant_i.h"
 #include "driver_i.h"
 
-#include "zephyr_fmac_main.h"
 #include "supp_main.h"
 #include "supp_api.h"
 
@@ -105,14 +104,6 @@ out:
 	}
 }
 
-int zephyr_supp_scan(const struct device *dev, scan_result_cb_t cb)
-{
-	const struct wifi_nrf_dev_ops *dev_ops = dev->api;
-
-	return dev_ops->off_api.disp_scan(dev, cb);
-}
-
-
 static inline void wpa_supp_restart_status_work(void)
 {
 	k_work_cancel_delayable_sync(&wpa_supp_status_work,
@@ -120,6 +111,7 @@ static inline void wpa_supp_restart_status_work(void)
 	k_work_reschedule(&wpa_supp_status_work,
 		K_MSEC(10));
 }
+
 
 int zephyr_supp_connect(const struct device *dev,
 						struct wifi_connect_req_params *params)
--- a/zephyr/src/supp_api.c
+++ b/zephyr/src/supp_api.c
@@ -1,0 +1,294 @@
+/**
+ * Copyright (c) 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/logging/log.h>
+#include <zephyr/kernel.h>
+
+#include "includes.h"
+#include "common.h"
+#include "common/defs.h"
+#include "wpa_supplicant/config.h"
+#include "wpa_supplicant_i.h"
+#include "driver_i.h"
+
+#include "zephyr_fmac_main.h"
+#include "supp_main.h"
+#include "supp_api.h"
+
+int cli_main(int, const char **);
+extern struct wpa_global *global;
+
+enum requested_ops {
+	CONNECT = 0,
+	DISCONNECT
+};
+
+#define OP_STATUS_POLLING_INTERVAL 1
+
+struct wpa_supp_api_ctrl {
+	const struct device *dev;
+	enum requested_ops requested_op;
+	int connection_timeout;
+	struct k_work_sync sync;
+};
+
+static struct wpa_supp_api_ctrl wpa_supp_api_ctrl;
+
+static void supp_shell_connect_status(struct k_work *work);
+
+static K_WORK_DELAYABLE_DEFINE(wpa_supp_status_work,
+		supp_shell_connect_status);
+
+static int send_wpa_supplicant_dummy_event(void)
+{
+	struct wpa_supplicant_event_msg msg = { 0 };
+
+	msg.ignore_msg = true;
+
+	return send_wpa_supplicant_event(&msg);
+}
+
+
+static inline struct wpa_supplicant * get_wpa_s_handle(const struct device *dev)
+{
+	struct wpa_supplicant *wpa_s = wpa_supplicant_get_iface(global, dev->name);
+	if (!wpa_s) {
+		wpa_printf(MSG_ERROR,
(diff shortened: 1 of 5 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -5,64 +5,36 @@
...
-	k_mbox_put(&wpa_supp_mbox,
-		   &send_msg,
-		   K_FOREVER);
+	struct wpa_supplicant_event_msg msg = { 0 };
+
+	msg.ctx = ctx;
+	msg.event = event;
+	if (data) {
+		msg.data = os_zalloc(sizeof(*data));
+		if (!data) {
```
sachinthegreen, line 28: Shouldn't this be if (!msg.data) ?

src/drivers/driver_zephyr.c
```diff
@@ -5,64 +5,36 @@
...
-			       500,
-			       wpa_supp_drv_mbox_msg_handler,
-			       NULL,
-			       NULL);
-
-	k_mbox_put(&wpa_supp_mbox,
-		   &send_msg,
-		   K_FOREVER);
+	struct wpa_supplicant_event_msg msg = { 0 };
+
```
sachinthegreen, line 23: Might be better to add a defensive check to see that ctx is non-NULL.

src/drivers/driver_zephyr.c
```diff
@@ -510,8 +506,11 @@ static int wpa_drv_zep_authenticate(void *priv,
 	
 	if_ctx->ssid_len = params->ssid_len;
 
+	curr_bss = wpa_bss_get(if_ctx->supp_if_ctx, params->bssid, params->ssid, params->ssid_len);
+
```
sachinthegreen, line 510: Should there be a NULL check here (in case of failure)?

zephyr/src/supp_api.c
```diff
@@ -0,0 +1,274 @@
...
+enum requested_ops {
+	CONNECT = 0,
+	DISCONNECT
+};
+
+struct wpa_supp_api_ctrl {
+	const struct device *dev;
+	int requested_op;
+	int connection_timeout;
+};
```
sachinthegreen, line 31: Can these be moved to supp_api.h?

zephyr/src/supp_api.c
```diff
@@ -0,0 +1,274 @@
...
+		/* Disconnect is a synchronous operation i.e., we are already disconnected
+		 * we are just using this thread to post net_mgmt event
+		 */
+		wifi_mgmt_raise_disconnect_result_event(net_if_lookup_by_dev(ctrl->dev), 0);
+	}
+}
+
+K_THREAD_DEFINE(wpa_supp_api,
+				1024,
+				supp_shell_connect_status,
```
sachinthegreen, line 90: Indentation seems to be off.

zephyr/src/supp_main.c
```diff
@@ -13,19 +13,31 @@
...
 #include "p2p_supplicant.h"
 #include "wpa_supplicant_i.h"
+#include "driver_i.h"
+
+#include "supp_main.h"
+
+
+#define DUMMY_SOCKET_PORT 9999
+
+#define DUMMY_SOCKET_PORT 9999
```
sachinthegreen, line 37: Double define?

sachinthegreen (changes requested):

zephyr/src/supp_api.c
krish2718: These are only used in this C file.

src/drivers/driver_zephyr.c
krish2718: `ctx` cannot be NULL if we are in the `driver` layer.

zephyr/src/supp_api.c
```diff
@@ -0,0 +1,289 @@
...
+	DISCONNECT
+};
+
+struct wpa_supp_api_ctrl {
+	const struct device *dev;
+	enum requested_ops requested_op;
+	int connection_timeout;
+};
+
+struct wpa_supp_api_ctrl wpa_supp_api_ctrl;
```
rlubos, line 34: I believe this should be `static`.

zephyr/src/supp_api.c
```diff
@@ -0,0 +1,289 @@
...
+		goto out;
+	}
+
+	if (ctrl->requested_op == CONNECT) {
+		/* TODO: use a timer for connection timeout*/
+		while (wpa_s->wpa_state != WPA_COMPLETED &&
+			   (ctrl->connection_timeout == SYS_FOREVER_MS ||
+			    i++ < ctrl->connection_timeout))
+		{
+			k_yield();
```
rlubos, line 76: I don't think this yield is needed given that the thread is put to sleep just after.

zephyr/src/supp_api.c
```diff
@@ -0,0 +1,289 @@
...
+		wifi_mgmt_raise_connect_result_event(net_if_lookup_by_dev(ctrl->dev), status);
+	} else if (ctrl->requested_op == DISCONNECT) {
+		/* Disconnect is a synchronous operation i.e., we are already disconnected
+		 * we are just using this thread to post net_mgmt event
+		 */
+		wifi_mgmt_raise_disconnect_result_event(net_if_lookup_by_dev(ctrl->dev), 0);
+	}
+}
+
+K_THREAD_DEFINE(wpa_supp_api,
```
rlubos, line 99: Actually, I think it'd be better to use system work queue to check the connection status instead of starting a dedicated thread (and wasting 1k of RAM for the extra stack). You could encapsulate `struct k_work_delayable` within `struct wpa_supp_api_ctrl` and then access the ctrl structure with `CONTAINER_OF()` macro (it's a common practice in Zephyr).

Then, instead of putting the thread to sleep, just resubmit the work item to re-check the connection status after some time.

zephyr/src/supp_api.c
```diff
@@ -0,0 +1,289 @@
...
+	wpa_supplicant_select_network(wpa_s,
+				      ssid);
+
+	send_wpa_supplicant_dummy_event();
+
+	wpa_supp_api_ctrl.dev = dev;
+	wpa_supp_api_ctrl.requested_op = CONNECT;
+	wpa_supp_api_ctrl.connection_timeout = params->timeout;
+	/* Cleanup previous spawns */
+	k_thread_abort(wpa_supp_api);
```
rlubos, line 190: Generally thread aborting in Zephyr is not considered a safe operation (https://docs.zephyrproject.org/latest/kernel/services/threads/index.html#thread-aborting), especially that the kernel gives to guarantees that it'll be able to respawn the thread. That's another reason a delayable work might be a better fit here.

sachinthegreen (approved):

zephyr/src/supp_api.c
krish2718: Thread is better for shorter polls (`ms`), but as we only needs seconds granularity, we can use delayed workqueues as well.

rlubos (approved):

PR #21, opened 2022-09-08
Coverity fixes

krish2718:


zephyr/src/supp_main.c
```diff
@@ -89,6 +89,8 @@ static void iface_cb(struct net_if *iface, void *user_data)
 
 	/* TODO : make this user configurable*/
 	ifaces[0].ifname = "wlan0";
+
+	os_free(ifname);
```
sachinthegreen, line 93: We should have a check early on (line 83) to make sure that ifname is not NULL.

sachinthegreen (changes requested):

rlubos:
@sachinthegreen @sr1dh48r @krish2718 Please prioritize the review process, if we want to get this in we need to do this ASAP.

zephyr/src/supp_main.c
krish2718: os_free can take care of NULL pointer, but added a null check in the print.

sachinthegreen (approved):

sachinthegreen:
> @sachinthegreen @sr1dh48r @krish2718 Please prioritize the review process, if we want to get this in we need to do this ASAP.

@rlubos : Approved from my end. @sr1dh48r is on vacation.

rlubos (approved):

rlubos:
@krish2718 Can we merge and update SHA in https://github.com/nrfconnect/sdk-nrf/pull/8687?

PR #23, opened 2022-09-19
Wait for WPA supplicant initialization

krish2718:
Before processing wifi_mgmt APIs check and wait for WPA supplicant initialization. This solves bus fault when wifi_mgmt API is called too early (before WPA supplicant populates "global" pointer).

This is done using a binary semaphore with initial value as 1.

### DIFF
--- a/zephyr/src/supp_api.c
+++ b/zephyr/src/supp_api.c
@@ -18,6 +18,7 @@
 #include "supp_api.h"
 
 int cli_main(int, const char **);
+extern struct k_sem wpa_supplicant_ready_sem;
 extern struct wpa_global *global;
 
 enum requested_ops {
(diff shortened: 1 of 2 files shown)


sachinthegreen (approved):

rlubos (approved):

zephyr/src/supp_api.c
```diff
@@ -53,7 +54,17 @@ static int send_wpa_supplicant_dummy_event(void)
 
 static inline struct wpa_supplicant * get_wpa_s_handle(const struct device *dev)
 {
-	struct wpa_supplicant *wpa_s = wpa_supplicant_get_iface(global, dev->name);
+	struct wpa_supplicant *wpa_s = NULL;
+	int ret = k_sem_take(&wpa_supplicant_ready_sem, K_SECONDS(2));
```
wbober, line 58: Time value is quite arbitrary here. Can supplicant init fail? If not why not put `K_FOREVER`?

wbober (approved):

zephyr/src/supp_api.c
krish2718: It shouldn't, but don't want to block instead return failure, as blocking here would block shell.

PR #25, opened 2022-09-26
Add support for channel selection

krish2718:
If a channel is provided pass it WPA supplicant to use that in both scan and connection.

The conversion from channel to frequency without any regulatory information (country code, OP class) is tricky, but use Global with widest OP classes common to most countries (81 and 128).

### DIFF
--- a/zephyr/src/supp_api.c
+++ b/zephyr/src/supp_api.c
@@ -146,6 +146,38 @@ int zephyr_supp_connect(const struct device *dev,
 	ssid->disabled = 1;
 	ssid->key_mgmt = WPA_KEY_MGMT_NONE;
 
+	if (params->channel != WIFI_CHANNEL_ANY) {
+		/* We use global channel list here and also use the widest
+		 * op_class for 5GHz channels as there is no user input
+		 * for these.
+		 */
+		int freq  = ieee80211_chan_to_freq(NULL, 81, params->channel);
+
+		if (freq <= 0) {
+			freq  = ieee80211_chan_to_freq(NULL, 128, params->channel);
+		}
+
+		if (freq <= 0) {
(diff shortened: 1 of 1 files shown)


rado17 (approved):

zephyr/src/supp_api.c
```diff
@@ -146,6 +146,37 @@ int zephyr_supp_connect(const struct device *dev,
...
+		if (freq <= 0) {
+			freq  = ieee80211_chan_to_freq(NULL, 128, params->channel);
+		}
+
+		if (freq <= 0) {
+			wpa_printf(MSG_ERROR, "Invalid channel %d", params->channel);
+			return -EINVAL;
+		}
+
+		ssid->scan_freq = os_zalloc(2 * sizeof(int));
```
rlubos, line 165: Who frees the allocated memory?

zephyr/src/supp_api.c
```diff
@@ -146,6 +146,37 @@ int zephyr_supp_connect(const struct device *dev,
...
+		if (!ssid->scan_freq) {
+			return -ENOMEM;
+		}
+		ssid->scan_freq[0] = freq;
+		ssid->scan_freq[1] = 0;
+
+		ssid->freq_list = os_zalloc(2 * sizeof(int));
+		if (!ssid->freq_list) {
+			return -ENOMEM;
```
rlubos, line 174: In case of error, shouldn't the already allocated memory be freed? (i. e. `ssid->scan_freq`).

zephyr/src/supp_api.c
krish2718: This is freed when we free the SSID i.e., if another connection request it sent or when wpa_supplicant de-initializes.

rlubos (approved):

PR #29, opened 2022-10-06
Move MbedTLS configuration files from sdk-nrf

krish2718:
Choose necessary MbedTLS Kconfig symbols by default

MbedTLS is only used by sdk-hostap, choose the necessary Kconfig
options by default instead of relying on samples.

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -27,7 +27,23 @@ config WEP
 config WPA_SUPP_CRYPTO
     bool "Crypto support for WiFi"
     select WEP
+    select MBEDTLS
+    select NRF_SECURITY
+    select MBEDTLS_LEGACY_CRYPTO_C
+    select MBEDTLS_ECP_C
+    select MBEDTLS_CTR_DRBG_C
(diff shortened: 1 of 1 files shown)


zephyr/Kconfig
```diff
@@ -27,8 +27,25 @@ config WEP
 config WPA_SUPP_CRYPTO
     bool "Crypto support for WiFi"
     select WEP
+    select MBEDTLS
+    select NRF_SECURITY
+    select MBEDTLS_LEGACY_CRYPTO_C
+    select MBEDTLS_ECP_ALL_ENABLED
```
LuDuda, line 33: Just a question.. Do we need all ECP algorithms, or just `MBEDTLS_ECP_DP_SECP256R1_ENABLED` would be enough?

krish2718: You are right, we only need that, and that is the only curve which is enabled by default when we choose MBEDTLS_ECP_C, so, can remove this line altogether.

sachinthegreen (approved):

PR #31, opened 2022-10-10
Fix Zephyr includes with latest up-merge

krish2718:
Zephyr has deprecated including "zephyr" directory in includes, so, all references should mandatory use "zephyr" prefix.

### DIFF
--- a/zephyr/src/wpa_cli.c
+++ b/zephyr/src/wpa_cli.c
@@ -12,15 +12,15 @@
 #include <stdlib.h>
 #include <ctype.h>
 
-#include <zephyr/zephyr.h>
+#include <zephyr/kernel.h>
 #include <zephyr/shell/shell.h>
 #include <zephyr/init.h>
 
-#include <net/net_if.h>
-#include <net/wifi_mgmt.h>
-#include <net/net_event.h>
+#include <zephyr/net/net_if.h>
+#include <zephyr/net/wifi_mgmt.h>
+#include <zephyr/net/net_event.h>
 
-#include <src/utils/common.h>
+#include <zephyr/src/utils/common.h>
 #include <wpa_supplicant/config.h>
 #include <wpa_supplicant/wpa_supplicant_i.h>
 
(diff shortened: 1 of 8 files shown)


anangl (changes requested):
This one also needs to be updated:
https://github.com/nrfconnect/sdk-hostap/blob/4ee2a9c76e28dfd590f289960b44e7b100cb4e04/src/utils/common.c#L1306-L1307
and those:
https://github.com/nrfconnect/sdk-hostap/blob/4ee2a9c76e28dfd590f289960b44e7b100cb4e04/src/utils/os_zephyr.c#L8
https://github.com/nrfconnect/sdk-hostap/blob/4ee2a9c76e28dfd590f289960b44e7b100cb4e04/src/l2_packet/l2_packet_zephyr.c#L14-L21

krish2718:
Thanks @anangl I have not built and fixed all issues.

anangl (approved):

sachinthegreen (approved):

zephyr/src/wpa_cli.c
```diff
@@ -12,15 +12,15 @@
...
 
-#include <net/net_if.h>
-#include <net/wifi_mgmt.h>
-#include <net/net_event.h>
+#include <zephyr/net/net_if.h>
+#include <zephyr/net/wifi_mgmt.h>
+#include <zephyr/net/net_event.h>
 
-#include <src/utils/common.h>
+#include <zephyr/src/utils/common.h>
```
anangl, line 23: This is incorrect. There is no such file.

anangl (changes requested):

zephyr/src/wpa_cli.c
anangl: Oops, a bit to late...

krish2718: oops..will raise another PR :-(.

krish2718: Its not even giving a build warning, may be I just do a force push.

krish2718: Wait, we need to use CONFIG_WPA_CLI=y

PR #32, opened 2022-10-10
Ctrl iface zephyr

krish2718:
Use Eloop and sockets for WPA CLI
  
  In Linux WPA CLI is a standalone application that uses sockets to
  communicate with ctrl_iface (uses Eloop), in Zephyr we have replaced
  sockets with direct function calls, but this introduces concurrency as
  we don't go through Eloop.
  
  Implement Zephyr's version of WPA CLI and control interface using
  socketpair UNIX style sockets, here WPA CLI can be invoked as a standard
  function rather than an executable.
  
  Also, as WPA CLI is meant for internal usage, remove exposing of limited
  set of commands and instead make it a pass through to accept any
  command.
  

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -0,0 +1,3785 @@
+/*
```
maxd-nordic, line 1: This huge file should not be copied from `wpa_cli.c` - better find a way to make the original work.

krish2718: We cannot re-use the file as it was intended to be a separate application with interactive usage, we could split the wpa_cli commands in to a common file and then re-use but that would be make maintaining it difficult which is why I forked it.

krish2718: For now I have not split the common code in to a separate file, to keep upmerging simple.

sr1dh48r (approved):

wpa_supplicant/ctrl_iface.c
```diff
@@ -11484,7 +11484,6 @@ char * wpa_supplicant_ctrl_iface_process(struct wpa_supplicant *wpa_s,
 		return NULL;
 	}
 
-	memset(reply, '\0', reply_size);
```
wbober, line 11487: What's wrong with that `memset`?

wbober:
Good job! If my understanding is correct you're also reverting a number of changes that were introduced when porting to Zephyr but now aren't needed, hence the code will be more aligned with vanilla wpa supplicant.

Two comments to the structure for your consideration:
1) I think it would really help do define API naming convention for functions that are zephyr specific, public, and safe for any zephyr thread to call. There is a mixture of naming prefixes that, at least I, find hard to comprehend. 

You could, for example, use `z_wpa_*` prefix. You could then rename:

```
zephyr_supp_status -> z_wpa_supplicant_status
zephyr_supp_disconnect -> z_wpa_supplicant_disconnect
zephyr_supp_connect -> z_wpa_supplicant_connect
```

2) I would keep the new zephyr related code in `zephyr/src` for the time being. Easier to understand what is added as part of the zephyr port.

wpa_supplicant/wpa_cli_zephyr.h
```diff
@@ -0,0 +1,44 @@
...
+#include "includes.h"
+#include "common/cli.h"
+#include "common/wpa_ctrl.h"
+#include "utils/common.h"
+#include "utils/eloop.h"
+#include "utils/edit.h"
+#include "utils/list.h"
+#include "wpa_supplicant_i.h"
+
+int wpa_cli_zephyr_init(struct wpa_supplicant *wpa_s);
```
wbober, line 23: I would rename to `z_wpa_ctrl_init()`, `z_wpa_ctrl_deinit()` and `z_wpa_cli()`.

zephyr/src/supp_api.c
```diff
@@ -65,27 +69,155 @@ static int send_wpa_supplicant_dummy_event(void)
...
+			break;
+		}
+		quote = make_argv(&cmd, c);
+	} while (true);
+
+	return quote;
+}
+
+
+int call_wpa_cli (const char *fmt, ...)
```
wbober, line 197: Should this be part of the `wpa_cli_zephyr.h` API as, for example, `z_wpa_cli_cmd_v`?

maxd-nordic: I would appreciate something like this to be exposed for ease of use. :)

zephyr/src/supp_api.c
```diff
@@ -165,100 +298,48 @@ int zephyr_supp_connect(const struct device *dev,
...
-			ssid->key_mgmt = WPA_KEY_MGMT_SAE;
-			str_clear_free(ssid->sae_password);
-			ssid->sae_password = dup_binstr(params->psk, params->psk_length);
-
-			if (ssid->sae_password == NULL) {
-				wpa_printf(MSG_ERROR, "%s:Failed to copy sae_password\n",
-					      __func__);
-				ret = -ENOMEM;
-				goto out;
+			if (params->sae_password) {
```
maxd-nordic, line 314: Don't forget to set key_mgmt: `call_wpa_cli("set_network %d key_mgmt SAE", resp.network_id);`

wpa_supplicant/wpa_cli_zephyr.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#ifndef __WPA_CLI_ZEPHYR_H_
+#define __WPA_CLI_ZEPHYR_H_
+
+#include <sys/unistd.h>
+
+#include <zephyr/kernel.h>
+
+#include "includes.h"
```
maxd-nordic, line 14: These internal includes are problematic. They should not be included in a header the app might try to use.

zephyr/src/supp_api.c
```diff
@@ -65,27 +69,155 @@ static int send_wpa_supplicant_dummy_event(void)
...
+
+	return quote;
+}
+
+
+int call_wpa_cli (const char *fmt, ...)
+{
+	char cmd[MAX_CMD_SIZE];
+	va_list cmd_args;
+	int ret;
```
maxd-nordic, line 201: unused now

zephyr/src/supp_api.c
```diff
@@ -149,14 +281,15 @@ static inline void wpa_supp_restart_status_work(void)
...
-
 int zephyr_supp_connect(const struct device *dev,
 						struct wifi_connect_req_params *params)
 {
 	struct wpa_ssid *ssid = NULL;
 	bool pmf = true;
 	struct wpa_supplicant *wpa_s;
 	int ret = 0;
+	struct add_network_resp resp = {0};
+	char buf[MAX_CMD_SIZE];
```
maxd-nordic, line 292: unused now

zephyr/src/supp_api.c
```diff
@@ -149,14 +281,15 @@ static inline void wpa_supp_restart_status_work(void)
 		K_MSEC(10));
 }
 
-
 int zephyr_supp_connect(const struct device *dev,
 						struct wifi_connect_req_params *params)
 {
 	struct wpa_ssid *ssid = NULL;
```
maxd-nordic, line 287: unused now

zephyr/src/supp_api.c
```diff
@@ -165,100 +298,48 @@ int zephyr_supp_connect(const struct device *dev,
...
 
-	wpa_supplicant_enable_network(wpa_s,
-				      ssid);
-
-	wpa_supplicant_select_network(wpa_s,
-				      ssid);
+	/* enable and select network */
+	call_wpa_cli("enable_network %d", resp.network_id);
 
-	send_wpa_supplicant_dummy_event();
```
maxd-nordic, line 261: send_wpa_supplicant_dummy_event is completely unused now

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -0,0 +1,134 @@
...
+ * See README for more details.
+ */
+
+#include "ctrl_iface_zephyr.h"
+
+static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
+					      void *sock_ctx)
+{
+	struct wpa_supplicant *wpa_s = eloop_ctx;
+	struct ctrl_iface_priv *priv = sock_ctx;
```
maxd-nordic, line 15: unused now

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -0,0 +1,3891 @@
...
+	ret = wpa_ctrl_command(ctrl, buf);
+
+out:
+	if (buf)
+		os_free(buf);
+
+	return ret;
+}
+
+static int wpa_cli_cmd_resp(struct wpa_ctrl *ctrl, const char *cmd, int min_args,
```
maxd-nordic, line 135: unused

glarsennordic:
Chiming in, I'm reviewing a related PR,

https://github.com/nrfconnect/sdk-nrf/pull/9261

Right now that PR is forced to make heavy use of `call_wpa_cli`, but that usage could (in my opinion) be reduced significantly if `zephyr_supp_connect` were broken down into smaller sub-functions, related to each step of setting up and then associating with a network.

I would like to suggest that `zephyr_supp_connect` be broken down in that way.

Perhaps there should even be a callable function for each individual command the CLI accepts?

zephyr/src/supp_api.c
```diff
@@ -65,27 +69,155 @@ static int send_wpa_supplicant_dummy_event(void)
...
+	return str == NULL ? 0U : (uint16_t)strlen(str);
+}
 
-static inline struct wpa_supplicant * get_wpa_s_handle(const struct device *dev)
+static char make_argv(char **ppcmd, uint8_t c)
 {
-	struct wpa_supplicant *wpa_s = NULL;
-	int ret = k_sem_take(&wpa_supplicant_ready_sem, K_SECONDS(2));
+	char *cmd = *ppcmd;
+	char quote = 0;
```
glarsennordic, line 81: AFAICT, quote will never have any value other than 0 in this function, is that intended?

krish2718: Yes, this can be removed.

krish2718:
> Chiming in, I'm reviewing a related PR,
> 
> [nrfconnect/sdk-nrf#9261](https://github.com/nrfconnect/sdk-nrf/pull/9261)
> 
> Right now that PR is forced to make heavy use of `call_wpa_cli`, but that usage could (in my opinion) be reduced significantly if `zephyr_supp_connect` were broken down into smaller sub-functions, related to each step of setting up and then associating with a network.
I am not sure I understand how breaking into smaller functions reduces `call_wpa_cli`, since we need to call `call_wpa_cli` for each of them anyway, with this new interface WPA supplicant only offers individual settings.
> 
> I would like to suggest that `zephyr_supp_connect` be broken down in that way.
> 
> Perhaps there should even be a callable function for each individual command the CLI accepts?
Well, sure, but we are trying to aggregate and expose a smaller set to the applications, we already have `wpa_cli` (disabled by default and can be enabled by `CONFIG_WPA_CLI=y` that exposes full commands of WPA supplicant.)
> 
> See this comment: [nrfconnect/sdk-nrf#9261 (comment)](https://github.com/nrfconnect/sdk-nrf/pull/9261#discussion_r1080712094)

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -0,0 +1,3891 @@
...
+{
+	int ret;
+	char buf[MAX_RESPONSE_SIZE] = {0};
+
+	ret = wpa_ctrl_command_resp(ctrl_conn, "STATUS", buf);
+	if (ret) {
+		return ret;
+	}
+
+	ret = sscanf((const char *)buf, "bssid=%*\nfreq=%*\nssid=%s", resp->ssid);
```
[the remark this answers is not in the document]
krish2718: `%*` is used to ignore the values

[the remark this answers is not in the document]
krish2718: had to escape % `%%*`

krish2718:
> These would still be higher-level than direct wpa_cli commands. I think since it is such a common design pattern to add a network before actually connecting, exposing those two as separate steps (and also including a way to remove/deselect networks) could make sense

Sure, but the APIs in `zephyr/src/supp_api.c` directly map to the Zephyr's wifi_mgmt offload APIs, we can divide the APIs in case you want it to used for other usecases, but for now wifi_mgmt is the only case.

wpa_supplicant/wpa_cli_zephyr.c
krish2718: I have bitten the bullet and did this.

zephyr/src/supp_api.c
krish2718: As its a copy, I left the code unchanged.

zephyr/src/supp_api.c
```diff
@@ -141,13 +116,12 @@ static inline void wpa_supp_restart_status_work(void)
 }
 
 
-int zephyr_supp_connect(const struct device *dev,
+int z_wpa_supplicant_connect(const struct device *dev,
```
maxd-nordic, line 119: is there a corresponding change in sdk-zephyr?

krish2718: Yes, the `sdk-nrf` PR linked has the manifest change.

maxd-nordic: :+1:

glarsennordic:
> Sure, but the APIs in zephyr/src/supp_api.c directly map to the Zephyr's wifi_mgmt offload APIs, we can divide the APIs in case you want it to used for other usecases, but for now wifi_mgmt is the only case.

I would be inclined to do so, but I also recognize that's my personal opinion. I'll leave the final decision up to you and @maxd-nordic and the other reviewers

krish2718:
To fix CI where one of the Matter sample was failing, disabled all non-zephyr commands by default using CONFIG_WPA_CLI, this saves about 2% flash.

wbober (approved):

glarsennordic (approved):

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -0,0 +1,133 @@
...
+		return;
+	}
+	buf[res] = '\0';
+
+	pos = buf;
+	while (*pos == ' ')
+		pos++;
+
+	reply = wpa_supplicant_ctrl_iface_process(wpa_s, pos,
+							&reply_len);
```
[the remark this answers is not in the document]
krish2718: The original code has this on the stack, but to reduce stack sizes moved this to HEAP.

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -0,0 +1,133 @@
...
+struct ctrl_iface_global_priv *
+wpa_supplicant_global_ctrl_iface_init(struct wpa_global *global)
+{
+	return (void *) 1;
+}
+
+void
+wpa_supplicant_global_ctrl_iface_deinit(struct ctrl_iface_global_priv *priv)
+{
+}
```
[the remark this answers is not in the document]
krish2718: We need to implement these dummy because they will be called from the generic control interface code, but for now these are no-ops.

wpa_supplicant/wpa_cli.c
```diff
@@ -21,3186 +20,211 @@
...
+	if (ifname == NULL)
+		return -1;
 
+#ifdef ANDROID
+	if (access(ctrl_iface_dir, F_OK) < 0) {
+		cfile = os_strdup(ifname);
+		if (cfile == NULL)
+			return -1;
+	}
+#endif /* ANDROID */
```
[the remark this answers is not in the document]
krish2718: Yes, keeping changes minimal for easier upmerge.

wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -0,0 +1,3692 @@
...
+		goto out;
+	}
+	ret = wpa_ctrl_command(ctrl, buf);
+
+out:
+	if (buf)
+		os_free(buf);
+
+	return ret;
+}
```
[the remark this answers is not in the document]
krish2718: Yes, but we don't use `wpa_cli.c` in Zephyr.

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -0,0 +1,339 @@
...
+/* Lifted from zephyr shell_utils.c to handle escapes */
+static inline uint16_t supp_strlen(const char *str)
+{
+	return str == NULL ? 0U : (uint16_t)strlen(str);
+}
+
+static char make_argv(char **ppcmd, uint8_t c)
+{
+	char *cmd = *ppcmd;
+	char quote = 0;
```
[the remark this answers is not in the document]
krish2718: Again, as this is a copy of the Zephyr's code, avoiding local changes.

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -0,0 +1,133 @@
...
+ * WPA Supplicant / Zephyr socket pair -based control interface
+ * Copyright (c) 2022, Jouni Malinen <j@w1.fi>
+ *
+ * This software may be distributed under the terms of the BSD license.
+ * See README for more details.
+ */
+
+#include "ctrl_iface_zephyr.h"
+
+static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
```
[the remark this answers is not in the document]
krish2718: These are internal functions, so, do we need to add docstrings?

wpa_supplicant/wpa_cli.c
```diff
@@ -21,3186 +20,211 @@
...
+static char ** wpa_list_cmd_list(void);
+static void update_creds(struct wpa_ctrl *ctrl);
+static void update_networks(struct wpa_ctrl *ctrl);
+static void update_stations(struct wpa_ctrl *ctrl);
+static void update_ifnames(struct wpa_ctrl *ctrl);
 
 
-static int wpa_cli_cmd_dpp_listen(struct wpa_ctrl *ctrl, int argc,
-				  char *argv[])
+static void usage(void)
```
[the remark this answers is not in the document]
krish2718: No, we don't use `wpa_cli.c` in Zephyr.

wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -0,0 +1,3692 @@
...
+		cmd++;
+	}
+
+	if (count > 1) {
+		wpa_printf(MSG_INFO, "Ambiguous command '%s'; possible commands:", argv[0]);
+		cmd = wpa_cli_commands;
+		while (cmd->cmd) {
+			if (os_strncasecmp(cmd->cmd, argv[0],
+					   os_strlen(argv[0])) == 0) {
+				wpa_printf(MSG_INFO, " %s", cmd->cmd);
```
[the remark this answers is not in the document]
krish2718: Yes, this is to add space b/w all possible combinations, its from the upstream code.

sachinthegreen (approved):

wpa_supplicant/ctrl_iface_zephyr.c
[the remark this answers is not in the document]
glarsennordic: You're right, docstrings are probably not required, but some comments at least would be a good idea, just to help any prospective readers out

I won't block this PR on that, though

wpa_supplicant/ctrl_iface_zephyr.c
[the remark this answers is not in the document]
glarsennordic: I would recommend some comments indicating that

wpa_supplicant/ctrl_iface_zephyr.c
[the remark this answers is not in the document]
krish2718: > but some comments at least would be a good idea, just to help any prospective readers out

Helpful comments are strictly against OSS philosophy :)

wpa_supplicant/wpa_cli_zephyr.c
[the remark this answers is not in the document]
glarsennordic: Where is it copied from in Zephyr, and could it be included instead?

And is the always-zero quote also an attribute of this copy?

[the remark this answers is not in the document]
krish2718: https://github.com/nrfconnect/sdk-hostap/pull/32/files#diff-226f1a66d69db480c0d31c83165ca0adb98ec2c93cbe13f7abdc958b0ddd29f2R106 this comment points to Zephyr's original.

Now that I see, I had removed stripping off the escapes, you are right, this had a local change, will raise a PR to remove quote completely.

PR #37, opened 2022-10-18
802154 fixes

krish2718:
Fixed for Wi-Fi + Thread combo build.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -670,10 +670,6 @@ static int wpa_drv_zep_get_ssid(void *priv,
 
 	if_ctx = priv;
 
-	if (!if_ctx->ssid) {
-		return 0;
-	}
-
 	wpa_printf(MSG_INFO,
 		   "%s: SSID size: %d\n",
 		   __func__,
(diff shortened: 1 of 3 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -670,10 +670,6 @@ static int wpa_drv_zep_get_ssid(void *priv,
 
 	if_ctx = priv;
 
-	if (!if_ctx->ssid) {
-		return 0;
-	}
-
```
sachinthegreen, line 676: Won't the further code crash if we remove this check?

src/l2_packet/l2_packet_zephyr.c
```diff
@@ -169,8 +169,7 @@ l2_packet_init(const char *ifname, const u8 *own_addr, unsigned short protocol,
 	ll.sll_protocol = htons(protocol);
 
 	// FIXME: This should skip bind_default to ETH_P_ALL, but its not.
```
sachinthegreen, line 171: Is this FIXME still valid after this change?

src/drivers/driver_zephyr.c
krish2718: No, that is the warning, ssid is an array on the stack, it can never be NULL.

src/l2_packet/l2_packet_zephyr.c
krish2718: Yes, this code doesn't change any functional behaviour, just a unnecessary NULL check

sachinthegreen (approved):

rlubos (approved):

PR #39, opened 2022-10-25
Use the PMF setting from the user

krish2718:
Now that Zephyr supports taking PMF setting from the user, use that instead of hard-coding to optional.


rlubos (approved):

sr1dh48r (approved):
Do we even need the pmf check anymore ? seems like its always set to true ..

PR #40, opened 2022-10-28
Remove unncessary error message

krish2718:
If a 802.15.4 interface is present, then this is expected and not an error, so, just silently ignore.


sr1dh48r (approved):

frkv (approved):
LGTM

PR #44, opened 2022-11-25
wifi : Adds support for handling WIPHY capabilities, RX MLME frames 

sr1dh48r:
Adds support for handling WIPHY capabilities, RX MLME frames to fix the following MBO issues

[SHEL-859] : Not able to send BTM response for the BTM request frame from the AP.
[SHEL-888] : Supported Operating classes IE missing in the Assoc Request frame.
[SHEL-877] : DUT not responding to AP beacon request frame using case 3 values

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -12,6 +12,7 @@
 #include "eloop.h"
 #include "driver_zephyr.h"
 #include "supp_main.h"
+#include "common/ieee802_11_common.h"
 
 #define SCAN_TIMEOUT 30
 
@@ -202,7 +203,7 @@ static void wpa_drv_zep_event_mgmt_tx_status(struct zep_drv_if_ctx *if_ctx,
 	const struct ieee80211_hdr *hdr;
 	u16 fc;
 
-	wpa_printf(MSG_DEBUG, "nl80211: Frame TX status event");
+	wpa_printf(MSG_DEBUG, "wpa_supp: Frame TX status event");
 
 	hdr = (const struct ieee80211_hdr *) frame;
 	fc = le_to_host16(hdr->frame_control);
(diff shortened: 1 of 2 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,422 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+	int last_mode, last_chan_idx;
+	int failed;
+	u8 dfs_domain;
+};
+
+static void phy_info_freq_cfg(struct hostapd_hw_modes *mode,
+		struct hostapd_channel_data *chan,
+		struct wpa_supp_event_channel *chnl_info)
+{
+	u8 channel;
```
sachinthegreen, line 252: initialize to 0. Also insert a blank line between variable declaration and code.

src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,422 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+		chan->allowed_bw &= ~HOSTAPD_CHAN_WIDTH_40M;
+	if (chnl_info->wpa_supp_flags & WPA_SUPP_CHAN_FLAG_FREQUENCY_ATTR_NO_80MHZ)
+		chan->allowed_bw &= ~HOSTAPD_CHAN_WIDTH_80;
+	if (chnl_info->wpa_supp_flags & WPA_SUPP_CHAN_FLAG_FREQUENCY_ATTR_NO_160MHZ)
+		chan->allowed_bw &= ~HOSTAPD_CHAN_WIDTH_160;
+
+	if (chnl_info->wpa_supp_flags & WPA_SUPP_CHAN_DFS_CAC_TIME_VALID) {
+		chan->dfs_cac_ms = (chnl_info->wpa_supp_time);
+	}
+	/*Other elements are not present*/
```
sachinthegreen, line 286: Insert blank line. Also have space after /* and before */

src/drivers/driver_zephyr.c
sachinthegreen, line 252: Add defensive checks for input parameters.

sachinthegreen (changes requested):

src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,445 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+}
+
+static int phy_info_band_cfg(struct phy_info_arg *phy_info,
+		struct wpa_supp_event_supported_band *band_info)
+{
+	struct hostapd_hw_modes *mode;
+	int ret;
+
+	if (phy_info->last_mode != band_info->band)
+	{
```
sachinthegreen, line 405: Brackets on the same line as if statement.

src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,445 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+		mode = os_realloc_array(phy_info->modes,
+				*phy_info->num_modes + 1,
+				sizeof(*mode));
+		if (!mode)
+		{
```
sachinthegreen, line 410: Brackets on the same line as if statement.

src/drivers/driver_zephyr.h
```diff
@@ -15,6 +15,82 @@
...
+	struct wpa_supp_event_vht_mcs_info vht_mcs;
+} __WPA_SUPP_PKD;
+
+struct wpa_supp_event_supported_band {
+	unsigned short wpa_supp_n_channels;
+	unsigned short wpa_supp_n_bitrates;
+#define WPA_SUPP_SBAND_MAX_CHANNELS 29
+#define WPA_SUPP_SBAND_MAX_RATES 13
+	struct wpa_supp_event_channel channels[29];
+	struct wpa_supp_event_rate bitrates[13];
```
sachinthegreen, line 88: Use the macros above.

sachinthegreen (changes requested):

src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,444 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+	chan->freq = chnl_info->center_frequency;
+	chan->flag = 0;
+	chan->allowed_bw = ~0;
+	chan->dfs_cac_ms = 0;
+	
+	if (ieee80211_freq_to_chan(chan->freq, &channel) != NUM_HOSTAPD_MODES) {
+		chan->chan = channel;
+	}
+	if (chnl_info->wpa_supp_flags & WPA_SUPP_CHAN_FLAG_FREQUENCY_DISABLED)
+		chan->flag |= HOSTAPD_CHAN_DISABLED;
```
krish2718, line 263: Mix and match of braces for single if-condition, we should stick to either Zephyr (on) or Linux (off).

src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,444 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+		mode->vht_mcs_set[1] = 0xff;
+		mode->vht_mcs_set[4] = 0xff;
+		mode->vht_mcs_set[5] = 0xff;
+
+		*(phy_info->num_modes) += 1;
+
+		phy_info->last_mode = band_info->band;
+		phy_info->last_chan_idx = 0;
+	}
+	else
```
krish2718, line 439: balance the braces

src/drivers/driver_zephyr.c
```diff
@@ -236,6 +237,444 @@ static void wpa_drv_zep_event_proc_unprot_disassoc(struct zep_drv_if_ctx *if_ctx
...
+	const struct ieee80211_mgmt *mgmt;
+
+	union wpa_event_data event;
+	u16 fc, stype;
+	int rx_freq = 0;
+
+	wpa_printf(MSG_MSGDUMP, "wpa_supp: Frame event");
+	mgmt = (const struct ieee80211_mgmt *)frame;
+
+	if (frame_len < 24) {
```
krish2718, line 519: better to use a sizeof() if we have one of a define

krish2718 (dismissed):
LGTM, but as we are copying some driver_nl80211* functions, was wondering if we should make effort now to refactor some code? e.g, `wpa_driver_nl80211_postprocess_modes`/`phy_info_band` etc. have no `drv` dependency, so, can be moved to `driver_common.c`

src/drivers/driver_zephyr.c
sr1dh48r: Its better to stick with supplicant formatting since we would have to upstream at some point.

src/drivers/driver_zephyr.c
sr1dh48r: Again, sticking to nl80211 logic for now since sizeof is much larger than 24.

sr1dh48r:
> LGTM, but as we are copying some driver_nl80211* functions, was wondering if we should make effort now to refactor some code? e.g, `wpa_driver_nl80211_postprocess_modes`/`phy_info_band` etc. have no `drv` dependency, so, can be moved to `driver_common.c`

While its generic, I don't see any reason for another driver to even use this since they might be able to query it in their own way.

sr1dh48r:
@sachinthegreen @krish2718 can you guys approve this !!

src/drivers/driver_zephyr.c
krish2718: Yes, in that case we can remove braces for single if.

krish2718:
> since they might be able to query it in their own way much like the other drivers currently are.

Those functions I was referring were converting data to hostapd structures not querying, so, can be re-used. But we can do it later as well.

sachinthegreen (approved):

PR #46, opened 2022-11-29
Fix handling connection timeout equal to 0

Damian-Nordic:
When the connection timeout was 0 the code would always fail the connection. Change the condition to handle the timeout only if it is positive.
By the way, simplify the conditions and unlock the mutex after modifying the shared variable.

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>

### DIFF
--- a/zephyr/src/supp_api.c
+++ b/zephyr/src/supp_api.c
@@ -108,25 +108,17 @@ static void supp_shell_connect_status(struct k_work *work)
 		goto out;
 	}
 
-	if (ctrl->requested_op == CONNECT) {
-		if (wpa_s->wpa_state != WPA_COMPLETED &&
-			   (ctrl->connection_timeout == SYS_FOREVER_MS ||
-			    seconds_counter++ <= ctrl->connection_timeout))
-		{
-			k_work_reschedule(&wpa_supp_status_work,
-				K_SECONDS(OP_STATUS_POLLING_INTERVAL));
-			k_mutex_unlock(&wpa_supplicant_mutex);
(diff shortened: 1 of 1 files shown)


zephyr/src/supp_api.c
```diff
@@ -108,25 +108,17 @@ static void supp_shell_connect_status(struct k_work *work)
...
-			   (ctrl->connection_timeout == SYS_FOREVER_MS ||
-			    seconds_counter++ <= ctrl->connection_timeout))
-		{
-			k_work_reschedule(&wpa_supp_status_work,
-				K_SECONDS(OP_STATUS_POLLING_INTERVAL));
-			k_mutex_unlock(&wpa_supplicant_mutex);
-			ctrl->status_thread_state = STATUS_THREAD_RUNNING;
-			return;
+	if (ctrl->requested_op == CONNECT && wpa_s->wpa_state != WPA_COMPLETED) {
+		if (ctrl->connection_timeout > 0 && seconds_counter++ <= ctrl->connection_timeout) {
```
sachinthegreen, line 112: Shouldn't this be seconds_counter++ **>** ctrl->connection_timeout ?

sachinthegreen (changes requested):

zephyr/src/supp_api.c
Damian-Nordic: Yes, of course!

sachinthegreen (approved):

LuDuda (approved):

PR #47, opened 2022-12-02
SAE-H2E MbedTLS fixes

krish2718:
MbedTLS uses an undocumented optimization where 0/null is used for A for all NIST curves to avoid doing `-3 mod P` calculation (saves a few computations).

So, instead of returning A as is, derive it using `-3 mod P` to handle the optimization.

### DIFF
--- a/src/crypto/crypto_mbedtls-ec.c
+++ b/src/crypto/crypto_mbedtls-ec.c
@@ -358,9 +358,10 @@ int crypto_ec_point_solve_y_coord(
     struct crypto_ec *e, struct crypto_ec_point *p,
     const struct crypto_bignum *x, int y_bit)
 {
-	mbedtls_mpi temp;
+	mbedtls_mpi temp, temp_no_alias;
 	mbedtls_mpi *y_sqr, *y;
 	mbedtls_mpi_init(&temp);
+	mbedtls_mpi_init(&temp_no_alias);
 	int ret = 0;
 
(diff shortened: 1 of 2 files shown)


src/crypto/crypto_mbedtls-ec.c
```diff
@@ -41,7 +43,19 @@ int crypto_rng_wrapper(void *ctx, unsigned char *buf, size_t len)
 
 const struct crypto_bignum *crypto_ec_get_a(struct crypto_ec *e)
 {
-	return (const struct crypto_bignum *)&e->group.A;
+	int ret;
```
sachinthegreen, line 46: ret doesn't seem to be getting used?

sachinthegreen, line 46: ret doesn't seem to be getting used?

sachinthegreen, line 46: ret doesn't seem to be getting used?

sachinthegreen (changes requested):

src/crypto/crypto_mbedtls-ec.c
krish2718: It's used by the macro `MBEDTLS_MPI_CHK` this needs both `ret` and `cleanup` to be defined.

sachinthegreen (approved):

sr1dh48r (approved):

PR #48, opened 2022-12-12
Add WEP support

krish2718:
Missed advertizing driver encryption capabilities to the WPA supplicant, so, it disallows WEP.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -1062,11 +1062,8 @@ static int _wpa_drv_zep_set_key(void *priv,
 		wpa_printf(MSG_ERROR, "%s: Invalid handle\n", __func__);
 		goto out;
 	}
-	if ((alg != WPA_ALG_NONE) &&
-	    ((!seq) || (!key))) {
-		wpa_printf(MSG_ERROR,
-			   "%s: Missing mandatory params\n",
-			   __func__);
+	if (alg != WPA_ALG_NONE && !key) {
+		wpa_printf(MSG_ERROR, "%s: Missing key\n", __func__);
 		goto out;
 	}
 
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -255,7 +255,7 @@ static void phy_info_freq_cfg(struct hostapd_hw_modes *mode,
 	chan->flag = 0;
 	chan->allowed_bw = ~0;
 	chan->dfs_cac_ms = 0;
-	
+
 	if (ieee80211_freq_to_chan(chan->freq, &channel) != NUM_HOSTAPD_MODES) {
 		chan->chan = channel;
 	}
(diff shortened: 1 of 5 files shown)


wpa_supplicant/wpa_cli.c
```diff
@@ -307,14 +307,20 @@ static int wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd)
...
     repl_buf = wpa_supplicant_ctrl_iface_process (wpa_s, (char *) cmd, &len);
-    
-	if (len > 0 && os_strcmp(repl_buf, "OK") != 0)
+
+	if (len == 2 && os_strcmp(repl_buf, "OK") != 0)
 	   printf ("Reply: %s\n", repl_buf);
-	
-	if (len > 0 && repl_buf)
-	    os_free(repl_buf);
+	else if (len == 5 && os_strcmp(repl_buf, "FAIL") == 0) {
```
sachinthegreen, line 313: Should len be 4 here?

sachinthegreen (changes requested):

wpa_supplicant/wpa_cli.c
krish2718: the string is `FAIL\n` so it should be 5, we are comparing only first 4, I can convert this to 5 and add \n.

sachinthegreen: Ok..I guess then the "OK" case above also needs to have a similar change?

sachinthegreen (changes requested):

src/drivers/driver_zephyr.c
```diff
@@ -1131,6 +1131,14 @@ static int wpa_drv_zep_get_capa(void *priv,
...
 	capa->rrm_flags |= WPA_DRIVER_FLAGS_SUPPORT_RRM;
 
+	capa->enc |= WPA_DRIVER_CAPA_ENC_WEP40 |
+			WPA_DRIVER_CAPA_ENC_WEP104 |
+			WPA_DRIVER_CAPA_ENC_TKIP |
+			WPA_DRIVER_CAPA_ENC_CCMP |
+			WPA_DRIVER_CAPA_ENC_CCMP |
+			WPA_DRIVER_CAPA_ENC_CCMP_256 |
+			WPA_DRIVER_CAPA_ENC_GCMP_256;
+
```
sachinthegreen, line 1141: Sorry for doing this in multiple passes..but I was just thinking....shouldn't this be gathered from the device driver (i.e. have a op defined which will return the capabilities which a particular device supports)?

sachinthegreen (changes requested):

src/drivers/driver_zephyr.c
krish2718: Absolutely, I did think of this and glanced but couldn't see any event from RPU, so, skipped it. But we can re-look or implement one, but may in another PR?

sachinthegreen: Yeah we presently don't have a way to get this information from the RPU but I was thinking for now lets atleast move this from driver_zephyr.c to zephyr_wpa_supp_if.c to make the wpa_supplicant code device agnostic. We can then have a TODO in zephyr_wpa_supp_if.c to get this from the RPU at a later stage.

sachinthegreen (approved):

sr1dh48r (approved):

PR #50, opened 2023-01-09
wifi : Bugfix for RRM, BTM request handling

sr1dh48r:
[SHEL-877]: DUT did not respond to AP beacon request frame
Enable RRM request/response frame handling.

[SHEL-859]: No BTM response for request from AP
Handle deauth during BSS transition

Signed-off-by: Sridhar Nuvusetty <sridhar.nuvusetty@nordicsemi.no>

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -507,6 +507,14 @@ static int wpa_drv_mgmt_subscribe_non_ap(struct zep_drv_if_ctx *if_ctx)
 	if (wpa_drv_register_action_frame(if_ctx, (u8 *)"\x0a\x07", 2) < 0)
 		ret = -1;
 
+	/* Radio Measurement - Neighbor Report Response */
+	if (wpa_drv_register_action_frame(if_ctx, (u8 *)"\x05\x05", 2) < 0)
+		ret = -1;
+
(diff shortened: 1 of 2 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -1121,7 +1129,6 @@ static int wpa_drv_zep_set_key(void* priv,
 static int wpa_drv_zep_get_capa(void *priv,
 			       	struct wpa_driver_capa *capa)
 {
-
```
krish2718, line 1124: unnecessary whitespace

src/utils/os_zephyr.c
```diff
@@ -45,6 +45,8 @@ int os_get_reltime(struct os_reltime *t)
 	struct timespec ts;
 	int res;
 
+	clock_id = CLOCK_REALTIME;
```
krish2718, line 48: This should be a separate commit with appropriate commit log

krish2718: FYI, I have root cause and raised https://github.com/nrfconnect/sdk-hostap/pull/53 with proper fix (added your SOB too). So, no need to switch to REALTIME, we can get it working with any clock setting.

sachinthegreen (approved):

PR #52, opened 2023-01-09
MAC address provisioning from host

sachinthegreen:
SHEL-1102 : MAC address provisioning from host
- Allow supplicant commands only if interface is up

Signed-off-by: Sridhar Nuvusetty <sridhar.nuvusetty@nordicsemi.no>

### DIFF
--- a/zephyr/src/supp_api.c
+++ b/zephyr/src/supp_api.c
@@ -158,6 +158,13 @@ int zephyr_supp_connect(const struct device *dev,
 	struct wpa_supplicant *wpa_s;
 	int ret = 0;
 
+	if (!net_if_is_up(net_if_lookup_by_dev(dev))) {
+		wpa_printf(MSG_ERROR,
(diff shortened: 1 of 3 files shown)


zephyr/src/supp_api.c
```diff
@@ -158,6 +158,13 @@ int zephyr_supp_connect(const struct device *dev,
 	struct wpa_supplicant *wpa_s;
 	int ret = 0;
 
+	if (!net_if_is_up(net_if_lookup_by_dev(dev))) {
```
krish2718, line 161: In case of iface NULL this throws an assert (if enabled), is it okay or should we check for it?

sr1dh48r (approved):

zephyr/src/supp_api.c
krish2718: I guess dev can never be NULL as we check in the init itself, so, clsoing.

PR #56, opened 2023-01-19
Improve delays

krish2718:
Instead of using sleep, switch to semaphore to avoid unnecessary extra loop.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -107,6 +107,7 @@ void wpa_drv_zep_event_proc_scan_done(struct zep_drv_if_ctx *if_ctx,
 			     if_ctx->supp_if_ctx);
 
 	if_ctx->scan_res2_get_in_prog = false;
+	k_sem_give(&if_ctx->drv_resp_sem);
 
 	wpa_supplicant_event_wrapper(if_ctx->supp_if_ctx,
 			EVENT_SCAN_RESULTS,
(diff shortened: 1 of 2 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -144,13 +146,13 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
...
-
-	return;
 err:
-	/* Ignore failures except the last */
 	if (!more_res) {
 		if_ctx->scan_res2_get_in_prog = false;
+		k_sem_give(&if_ctx->drv_resp_sem);
+	} else {
+		/* Ignore failures except the last */
+		if_ctx->scan_res2_get_in_prog = true;
```
sachinthegreen, line 155: Do we need the else part here?

src/drivers/driver_zephyr.c
```diff
@@ -662,16 +664,14 @@ struct hostapd_hw_modes *wpa_drv_get_hw_feature_data(void *priv,
...
-		k_yield();
-		os_sleep(1, 0);
-		i++;
+	if (ret < 0) {
+		return NULL;
 	}
 
+	k_sem_reset(&if_ctx->drv_resp_sem);
+	k_sem_take(&if_ctx->drv_resp_sem, K_SECONDS(GET_WIPHY_TIMEOUT));
+
```
sachinthegreen, line 674: Do we need error handling here if there is a timeout?

sachinthegreen (changes requested):

sachinthegreen (approved):

sr1dh48r (approved):

PR #60, opened 2023-02-06
Fix data access violation

krish2718:
If a sample doesn't use POSIX_API then the default POSIX_MAX_FDS are 4, but Zephyr's WPA supplicant needs at least 5, so, fix this by explicitly defining it.


sachinthegreen (approved):

rado17 (approved):

krish2718:
Modifies commits, please review again @sachinthegreen @rado17

PR #66, opened 2023-02-15
Add assert for minimum libc HEAP

krish2718:
This is to catch any RAM limitations early. But its a bit fiddly, as its runtime thing and board won't boot.

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -12,11 +12,25 @@ config WPA_SUPP
     select NET_SOCKETS_PACKET
     select NET_SOCKETPAIR
     select NET_L2_WIFI_MGMT
+    # For newlib min HEAP check, NCS_SAMPLE_DEFAULTS doesn't print assert information
+    select LOG_DEFAULT_MINIMAL
+	select ASSERT
+    select ASSERT_VERBOSE
+    # Without this reset on fatal error is triggered
+    select DEBUG
+    ###
     select EXPERIMENTAL if !SOC_NRF5340_CPUAPP_QKAA
     help
       WPA supplicant implements 802.1X related functions.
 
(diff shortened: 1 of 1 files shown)


sachinthegreen (approved):

rado17 (approved):

zephyr/Kconfig
```diff
@@ -12,12 +12,26 @@ config WPA_SUPP
...
+    # Without this reset on fatal error is triggered
+    select DEBUG
+    ###
     select EXPERIMENTAL if !SOC_NRF5340_CPUAPP_QKAA
     help
       WPA supplicant implements 802.1X related functions.
 
 if WPA_SUPP
 
+# For asserts, this avoids continuous reset
```
simensrostad, line 28: Should this be guarded by `CONFIG_ASSERT`? So that it doesn't become the default case when asserts are disabled, which will be the common case for most applications.

zephyr/Kconfig
```diff
@@ -12,12 +12,26 @@ config WPA_SUPP
...
+config RESET_ON_FATAL_ERROR
+    default n
+
+config NEWLIB_LIBC_MIN_REQUIRED_HEAP_SIZE
```
simensrostad, line 32: Just curious, why aren't we using zephyr system heap? How will this work if an application wants to use a differnt C library, PICOLIB for instance.

zephyr/Kconfig
krish2718: But we are selecting `ASSERT` above unconditionally, so, it would be redundant.

krish2718: I understand ASSERTS are not always desired but as of now that is the only to catch this low HEAP scenario gracefully.

PR #74, opened 2023-03-20
Add 7002 ns support

krish2718:


sachinthegreen (approved):

rado17 (approved):

src/crypto/crypto_mbedtls-ec.c
```diff
@@ -294,27 +297,29 @@ int crypto_ec_point_add(
 	return ret ? -1 : 0;
 }
 
+static int tls_ctr_drbg_random(void *ctx, unsigned char *buf, size_t len)
```
joerchan, line 300: Why is it is required to feed random into mbedtls in this way?
We should have the required integration to get entropy alread.

frkv: This is copying methodology in sockets_tls APIs in Zephyr. It is the same thing. TLS/DTLS  and legacy Mbed TLS APIs have frng/prng pointers they make use of.

frkv (approved):
LGTM

src/crypto/crypto_mbedtls-ec.c
joerchan: Ok, I can see that.

csrand may not be ctr_drbg. and sys_rand_get is not cryptographicaly secure.

I don't know if this is acceptable.

joerchan (approved):

src/crypto/crypto_mbedtls-ec.c
frkv: This is a copy-paste of the same function that sockets_tls.c is using.

PR #75, opened 2023-03-27
Fix POSIX build

krish2718:
As per [1], newlibc now defines _ANSI_SOURCE to avoid POSIX conflicts, but as hostapd needs POSIX stuff from libc, need to enable this explicitly.

[1] - https://github.com/zephyrproject-rtos/zephyr/issues/52739


rlubos (approved):

krish2718:
I thought this compiled successfully, but apparently not, so, please wait for proper fix.

sachinthegreen (approved):

PR #83, opened 2023-04-19
upmerge: Fix use after free

krish2718:


mbolivar-nordic:
Merged into https://github.com/nrfconnect/sdk-hostap/pull/82

PR #84, opened 2023-04-20
Add support to configure SAE-H2E

krish2718:
Before the control interface changes we had added a new command to set the H2E in the WPA supplicant configuration, but this was removed, so, add back the command that can be set through WPA cli.

### DIFF
--- a/wpa_supplicant/ctrl_iface.c
+++ b/wpa_supplicant/ctrl_iface.c
@@ -924,6 +924,17 @@ static int wpa_supplicant_ctrl_iface_set(struct wpa_supplicant *wpa_s,
 #endif /* CONFIG_WNM */
 	} else if (os_strcasecmp(cmd, "enable_dscp_policy_capa") == 0) {
 		wpa_s->enable_dscp_policy_capa = !!atoi(value);
+#ifdef CONFIG_SAE
+	} else if (os_strcasecmp(cmd, "sae_pwe") == 0) {
+		int pwe = atoi(value);
+
+		if (pwe < 0 || pwe > 2) {
+			wpa_printf(MSG_ERROR,
+				   "Invalid SAE PWE value %d", pwe);
+			return -1;
+		}
+		wpa_s->conf->sae_pwe = pwe;
+#endif /* CONFIG_SAE */
 	} else {
 		value[-1] = '=';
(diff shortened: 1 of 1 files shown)


wpa_supplicant/ctrl_iface.c
```diff
@@ -924,6 +924,17 @@ static int wpa_supplicant_ctrl_iface_set(struct wpa_supplicant *wpa_s,
 #endif /* CONFIG_WNM */
 	} else if (os_strcasecmp(cmd, "enable_dscp_policy_capa") == 0) {
 		wpa_s->enable_dscp_policy_capa = !!atoi(value);
+#ifdef CONFIG_SAE
+	} else if (os_strcasecmp(cmd, "sae_pwe") == 0) {
+		wpa_s->conf->sae_pwe = atoi(value);
+		if (wpa_s->conf->sae_pwe < 0 ||
+		    wpa_s->conf->sae_pwe > 2) {
```
rado17, line 931: Shouldn't the validity be checked before setting the parameter in wpa_s?

ajayparida (approved):

wpa_supplicant/ctrl_iface.c
krish2718: Thanks, fixed it.

wpa_supplicant/ctrl_iface.c
```diff
@@ -924,6 +924,17 @@ static int wpa_supplicant_ctrl_iface_set(struct wpa_supplicant *wpa_s,
 #endif /* CONFIG_WNM */
 	} else if (os_strcasecmp(cmd, "enable_dscp_policy_capa") == 0) {
 		wpa_s->enable_dscp_policy_capa = !!atoi(value);
+#ifdef CONFIG_SAE
+	} else if (os_strcasecmp(cmd, "sae_pwe") == 0) {
+		if (wpa_s->conf->sae_pwe < 0 ||
+		    wpa_s->conf->sae_pwe > 2) {
```
rado17, line 930: Validity check is required for value

rado17 (changes requested):

wpa_supplicant/ctrl_iface.c
krish2718: oops, my bad, fixed now.

rado17 (approved):

PR #86, opened 2023-04-27
Fix handling of IEs in authentication

krish2718:
When copying event from driver and postnig to WPA supplicant, the IE's are not copied, so, if WPA supplicant takes longer to process the event the driver frees the IEs causing invalid data to be processed by WPA supplicant. This is a common issue to all events.

For now Authentication events are being affected due to WPA3-SAE taking longer to process, so, fix only Authentication events for now as a generic fix needs more work and thought.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -32,6 +32,21 @@ void wpa_supplicant_event_wrapper(void *ctx,
 			return;
 		}
 		os_memcpy(msg.data, data, sizeof(*data));
+		if (event == EVENT_AUTH) {
+			union wpa_event_data *data_tmp = msg.data;
+
+			if (data->auth.ies) {
+				char *ies = os_zalloc(data->auth.ies_len);
+
+				if (!ies) {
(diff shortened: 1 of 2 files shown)


rado17 (approved):

src/drivers/driver_zephyr.c
```diff
@@ -157,6 +157,17 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
...
+	if (event->auth.ies) {
+		char *ies = os_zalloc(event->auth.ies_len);
+
+		if (!ies) {
+			wpa_printf(MSG_ERROR, "%s: Failed to alloc ies\n", __func__);
+			return;
+		}
+
+		os_memcpy(ies, event->auth.ies, event->auth.ies_len);
+		event->auth.ies = ies;
```
ajayparida, line 169: Replacing driver pointer wil a new memory will probably cause memory leak

sachinthegreen (approved):

PR #88, opened 2023-05-02
Add support for unsolicited scan results

krish2718:
For connected scan (from WPA supplicant) UMAC sends scan results on the fly i.e., it doesn't store and wait for WPA supplicant to issue get scan results.

Add support to accept and store these unsolicited scan results, they will still be returned to supplicant only when it asks.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -122,15 +122,24 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
 	struct wpa_scan_res **tmp = NULL;
 	size_t scan_res_len  = sizeof(struct wpa_scan_res) + r->ie_len + r->beacon_ie_len;
 
-	if (!if_ctx->scan_res2)
-		return;
+	if (!if_ctx->scan_res2) {
+		if_ctx->scan_res2 = os_zalloc(sizeof(*if_ctx->scan_res2));
+		wpa_printf(MSG_INFO, "%s: Alloc memory for scan results\n", __func__);
+	}
+
+	if (!if_ctx->scan_res2) {
+		wpa_printf(MSG_ERROR, "%s: Failed to alloc scan result\n", __func__);
+		goto err;
+	}
 
 	tmp = os_realloc_array(if_ctx->scan_res2->res,
 			       if_ctx->scan_res2->num + 1,
 			       sizeof(struct wpa_scan_res *));
 
 	if (!tmp) {
-		wpa_printf(MSG_ERROR, "%s: Failed to realloc scan result array\n", __func__);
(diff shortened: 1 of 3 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -122,15 +122,25 @@ void wpa_drv_zep_event_proc_scan_res(struct zep_drv_if_ctx *if_ctx,
...
 
-	if (!if_ctx->scan_res2)
-		return;
+	if (!if_ctx->scan_res2) {
+		if_ctx->scan_res2 = os_zalloc(sizeof(*if_ctx->scan_res2));
+		wpa_printf(MSG_INFO, "%s: Alloc memory for scan results\n", __func__);
+	}
+
+
+	if (!if_ctx->scan_res2) {
```
sachinthegreen, line 131: Can we move this check immediately after the os_zalloc?

src/drivers/driver_zephyr.c
```diff
@@ -821,6 +831,8 @@ static void *wpa_drv_zep_init(void *ctx,
 		goto out;
 	}
 
+	if_ctx->scan_res2 = NULL;
+
```
sachinthegreen, line 835: Do we need to do this (since os_zalloc is already being done for if_ctx above)?

src/drivers/driver_zephyr.c
```diff
@@ -938,7 +956,10 @@ struct wpa_scan_results *wpa_drv_zep_get_scan_results2(void *priv)
 		goto out;
 	}
 
-	if_ctx->scan_res2 = os_zalloc(sizeof(*if_ctx->scan_res2));
+	if (!if_ctx->scan_res2) {
+		if_ctx->scan_res2 = os_zalloc(sizeof(*if_ctx->scan_res2));
```
sachinthegreen, line 960: Can this be combined with the alloc check being done below ?

sachinthegreen (changes requested):

src/drivers/driver_zephyr.c
krish2718: This was intentional, if it's either non-NULL or allocation failure we need to catch that.

src/drivers/driver_zephyr.c
krish2718: This was intentional, if it's either non-NULL or allocation failure we need to catch that.

src/drivers/driver_zephyr.c
krish2718: No need, just an explicit assignment.

rado17 (approved):

krish2718:
Abandoning this as this is not working with `overlay-zperf.conf` needs more `libc malloc` memory.

PR #91, opened 2023-05-05
hostap: Read peer TWT capability

ajayparida:
Update peer TWT capability from connection info.


krish2718 (approved):
Please raise a sdk-nrf PR and update manifest for CI to pass.

PR #92, opened 2023-05-08
Use unbuffered logging for debugs

krish2718:
WPA supplicant debugs are excessive and the incoming rate is high esp. during connection, so, even though we increase the log buffer size, we still miss quite a few events.

So, disable buffered logging for debugs, one caveat is that we lose the timestamp that comes with LOG_ macros and also the prints might be jumbled but this is still preferred over losing events.


zephyr/Kconfig
```diff
@@ -115,4 +115,11 @@ config WPA_SUPP_DEBUG_LEVEL
       available levels and functions for emitting the messages. Note that
       runtime filtering can also be configured in addition to the compile-time
       filtering.
+
+config WPA_SUPP_DEBUG_SHOW_KEYS
```
sachinthegreen, line 119: Is it possible to make WPA_SUPP_LOG_LEVEL_DBG as a pre-condition for this?

sachinthegreen (approved):

rado17 (approved):

zephyr/Kconfig
markaj-nordic: +1

zephyr/src/supp_main.c
```diff
@@ -404,6 +404,14 @@ static void z_wpas_start(void)
 		goto out;
 	}
 
+#if defined (CONFIG_WPA_SUPP_LOG_LEVEL_DBG) && defined (CONFIG_WPA_SUPP_DEBUG_SHOW_KEYS)
+	wpa_supplicant_set_debug_params(global, MSG_DEBUG, 1, 1);
+#elif defined (CONFIG_WPA_SUPP_LOG_LEVEL_DBG)
+	wpa_supplicant_set_debug_params(global, MSG_DEBUG, 1, 0);
+#else
+	wpa_supplicant_set_debug_params(global, MSG_INFO, 0, 0);
```
markaj-nordic, line 412: So for the default we will end up with INFO. In Matter we set `WPA_SUPP_LOG_LEVEL_ERR`=y - is it going to be overridden by the code above?

zephyr/Kconfig
```diff
@@ -116,8 +116,10 @@ config WPA_SUPP_DEBUG_LEVEL
       runtime filtering can also be configured in addition to the compile-time
       filtering.
 
-# Debug logs need more buffer space
-config LOG_BUFFER_SIZE
-    default 4096 if WPA_SUPP_LOG_LEVEL_DBG
-    default 2048
+config WPA_SUPP_DEBUG_SHOW_KEYS
```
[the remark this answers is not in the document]
krish2718: Yes, in general `default n` is not allowed.

PR #93, opened 2023-05-10
Fix MbedTLS HEAP

krish2718:
Commit f13d1c9b95effd9e0b802bf8566348219ea0e725 ("nrf_security: Remove usage of HEAP in platform.c") broke Wi-Fi security as MbedTLS calloc/free are now undefined and any allocations by MbedTLS will fail resulting in crypto failures.

Fix this by setting alternatives to stdlib variants.

### DIFF
--- a/zephyr/src/supp_main.c
+++ b/zephyr/src/supp_main.c
@@ -16,6 +16,10 @@
 #include <sys/fcntl.h>
 LOG_MODULE_REGISTER(wpa_supplicant, LOG_LEVEL_DBG);
 
+#if defined(CONFIG_WPA_SUPP_CRYPTO) && !defined(CONFIG_MBEDTLS_ENABLE_HEAP)
+#include <mbedtls/platform.h>
+#endif /* CONFIG_WPA_SUPP_CRYPTO */
+
 #include "includes.h"
 #include "common.h"
 #include "eloop.h"
@@ -376,6 +380,12 @@ static void z_wpas_start(void)
 {
 	struct wpa_params params;
 	int exitcode = -1;
+
+#if defined(CONFIG_WPA_SUPP_CRYPTO) && !defined(CONFIG_MBEDTLS_ENABLE_HEAP)
+	/* Needed for crypto operation as default is no-op and fails */
+	mbedtls_platform_set_calloc_free(calloc, free);
+#endif /* CONFIG_WPA_CRYPTO */
+
 	k_work_queue_init(&z_wpas_iface_wq);
 
 	k_work_queue_start(&z_wpas_iface_wq,


rado17 (approved):

zephyr/src/supp_main.c
```diff
@@ -376,6 +380,12 @@ static void z_wpas_start(void)
 {
 	struct wpa_params params;
 	int exitcode = -1;
+
+#ifdef CONFIG_WPA_SUPP_CRYPTO
+	/* Needed for crypto operation as default is no-op and fails */
+	mbedtls_platform_set_calloc_free(calloc, free);
```
rlubos, line 386: I think this needs some clarifications from the `nrf_security` team, the commit mentioned here said:
> We always call mbedtls_platform_set_calloc_free to set alternative
heap management functions so this is unneccesary.

So it'd expect some reasonable calloc/free handler to be configured by the nrf_security lib. At least that'd make more sense to me, instead of having mbed TLS configuration buried inside some external library.

@joerchan Or perhaps I misunderstood the commit message of https://github.com/nrfconnect/sdk-nrfxlib/commit/f13d1c9b95effd9e0b802bf8566348219ea0e725? Can you shed some light on how heap management is supposed to be configured with `nrf_security`?

krish2718: I had the same doubt, but a quick search shows others (openthread, matter etc) are doing the same, initialising with their own or stdlib handlers, so, went with this approach.

LuDuda: @krish2718 I don't think this is accurate. At least for Nordic's platform neither OpenThread nor Matter calls `mbedtls_platform_set_calloc_free`. It is true you can find this function being called in those repos, but it does not apply to nRF Connect SDK.

I agree that it is not the best place to configure mbedtls general stuff in Wi-Fi module. At the same time, a hotfix is needed! :)

LuDuda: I think it might be the configuration issue on Wi-Fi side. Look here:

```
ludu@ludu-pc: nrf/samples/matter/lock$ cat build/zephyr/.config | grep "CONFIG_MBEDTLS_ENABLE_HEAP"
CONFIG_MBEDTLS_ENABLE_HEAP=y

ludu@ludu-pc: nrf/samples/wifi/shell$ cat build/zephyr/.config | grep "CONFIG_MBEDTLS_ENABLE_HEAP"
# CONFIG_MBEDTLS_ENABLE_HEAP is not set
```

Reference: https://github.com/nrfconnect/sdk-nrfxlib/blob/93d76b15567860b33ba5388fe6f0d3eb9592b3fe/nrf_security/src/zephyr/CMakeLists.txt#L47

krish2718: Wi-Fi doesn't enable MbedTLS HEAP, it is only enabled by default for TF-M, and it used to work till now, the change about calloc/free isn't dependent on MBEDTLS_HEAP.

sachinthegreen (approved):

zephyr/src/supp_main.c
LuDuda: So at least this call should be guarded by `#ifndef CONFIG_MBEDTLS_ENABLE_HEAP`. Else it overrides e.g. Matter default configuration in hidden way.

joerchan: @krish2718 
Sorry for breaking your stuff and not catching it.
Let's start with why this got broken. It appears that the sdk-hostap test-suite is not run when we are making changes to sdk-nrfxlib.
You can see here what was run: https://github.com/nrfconnect/sdk-nrf/pull/10990#issuecomment-1529606641

As for the reason why the change was made.
By having free and calloc be set as the default choice then these functions are always included, and we always end up with C stdlib heap functions. This is a problem for TF-M which does not want to include heap at all.

If you need to use heap you should have enabled CONFIG_MBEDTLS_ENABLE_HEAP. It has been working because of the always included calloc/heap functions. But these are now removed, revealing that you had a missing dependency.

For the fix, making sure that you have the dependency is what should be the correct fix, i.e CONFIG_MBEDTLS_ENABLE_HEAP should be enabled when you need heap.
If you don't want to enable this function for whatever reason then you can call mbedtls_platform_set_calloc_free to set your own.

However if you set your own you need to make sure that you are not interfering with other users of the heap.

Setting it unconditionally like this you may be overwriting what has already been set by other users, as @LuDuda points out you may be overriding Matter default configuration. Now we have to check the order of what is being called to find out what is being configured.

At the very least you should check `CONFIG_MBEDTLS_ENABLE_HEAP` before calling this function, otherwise you are actually changing Matter + Wi-Fi behavior with this change.

joerchan (changes requested):
Requesting changer for either enabling CONFIG_MBEDTLS_ENABLE_HEAP or at least check that it is enabled before setting calloc/free as the heap management functions.

joerchan:
@krish2718 It appears sdk-hostap is not mentioned in any of the test-spec.yml files. Your team should make an action to include your repository on crypto and nrf_security changes.
For example matter:
```
"CI-matter-test":
  - "crypto/**/*"
  - "nrf_security/**/*"
```

LuDuda (approved):
Thanks!

zephyr/src/supp_main.c
krish2718: Update with the recommended change to do this only when `CONFIG_MBEDTLS_ENABLE_HEAP` is not enabled.

krish2718:
> @krish2718 It appears sdk-hostap is not mentioned in any of the test-spec.yml files. Your team should make an action to include your repository on crypto and nrf_security changes. For example matter:
> 
> ```
> "CI-matter-test":
>   - "crypto/**/*"
>   - "nrf_security/**/*"
> ```

Done, see the `sdk-nrf` PR for the updates.

joerchan (approved):

PR #97, opened 2023-06-07
Process only Wi-Fi interface events

krish2718:
Use the pattern matching on interface names to filter only Wi-Fi interfaces.

### DIFF
--- a/zephyr/src/supp_main.c
+++ b/zephyr/src/supp_main.c
@@ -39,6 +39,7 @@ K_SEM_DEFINE(z_wpas_ready_sem, 0, 1);
 
 /* Should match with the driver name */
 #define DEFAULT_IFACE_NAME "wlan0"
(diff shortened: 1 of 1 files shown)


zephyr/src/supp_main.c
```diff
@@ -222,6 +223,11 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
 {
 	const char *ifname = iface->if_dev->dev->name;
 
+	if (strncmp(ifname, IFACE_MATCHING_PREFIX, strlen(IFACE_MATCHING_PREFIX)) != 0)
```
jtguggedal, line 226: Could use `sizeof(IFACE_MATCHING_PREFIX) - 1` instead of `strnlen` to reduce the runtime overhead

rado17 (approved):

sachinthegreen (approved):

PR #98, opened 2023-06-08
crypto: Remove remnants of mbedtls entropy usage

joerchan:
Remove remnants of mbedtls entropy usage.
Entropy is now accessed through zephyr APIs so enabling MBEDTLS_ENTROPY_C is no longer needed.
Without MBEDTLS_ENTROPY_C option there is no reason to change the default configuration value of MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG.

NCSDK-22096

### DIFF
--- a/src/crypto/tls_mbedtls.c
+++ b/src/crypto/tls_mbedtls.c
@@ -17,7 +17,6 @@
 #include <mbedtls/ssl.h>
 #include <mbedtls/x509_crt.h>
 #include <mbedtls/ctr_drbg.h>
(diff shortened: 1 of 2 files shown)


krish2718:
> entropy is added to a struct but only referenced when it is freed.
Yes, this is leftover code when we added support for PSA 7ad388cf1 removed the entropy except for these cleanups.

> Looks safe to delete for me.

Yes

sachinthegreen (approved):

frkv (approved):
LGTM

PR #106, opened 2023-07-07
zephyr: Kconfig: Fix PSA dependency on random numbers

joerchan:
Fix PSA dependency on random number support.
Random number generated is wanted, instead of a specific PRNG algorithm.


joerchan:
Note: Since this is only used with TF-M and  in that configuration PSA_WANT_GENERATE_RANDOM is included from the entropy driver for PSA this was not caught by CI. This may be a problem for boards that do not have this in their DTS file.

krish2718 (approved):

sachinthegreen (approved):

krish2718:
@joerchan you need to rebase this with latest hostapd for the Wi-Fi operations to work and fix CI failures.

joerchan:
> @joerchan you need to rebase this with latest hostapd for the Wi-Fi operations to work and fix CI failures.

Thanks. It passed now.

PR #108, opened 2023-07-10
Include BSS max idle timeout in association request

rado17:
As per 802.11-2020: 11.21.13 BSS max idle period management

"If dot11WirelessManagementImplemented is true,
dot11BSSMaxIdlePeriod is nonzero and
dot11BSSMaxIdlePeriodIndicationByNonAPSTA is true, then a non-S1G non-AP STA shall include a BSS Max Idle Period element in the (Re)Association Request frame. If the BSS Max Idle Period element is present in the (Re)Association Request frame received by a non-S1G AP that has dot11BSSMaxIdlePeriodIndicationByNonAPSTA equal to true, then the non-S1G AP may choose the non-AP STA’s preferred maximum idle period. The non-S1G AP indicates its chosen value to the non-S1G STA in the (Re)Association Response frame."

Add BSS max idle timeout (default 300s) in association request. Use CONFIG_BSS_MAX_IDLE_TIME to modify this timeout at build time.


zephyr/Kconfig
```diff
@@ -118,6 +118,15 @@ config POSIX_MAX_FDS
     # z_wpa_event_sock - 1 socketpair = 2
     default 7 if !POSIX_API
 
+config BSS_MAX_IDLE_TIME
+    int "BSS max idle timeout in seconds"
+    range 1 64000
```
krish2718, line 123: how do we disable this? Shouldn't we have 0 included?

krish2718 (commented):
Should we do this upstream and do a `fromlist` ?

zephyr/Kconfig
rado17: Agree.

krish2718 (approved):

sachinthegreen (approved):

VivekUppunda (approved):

PR #109, opened 2023-07-13
Supplicant changes to run upstream samples

rlubos:
* 39209a3e0ef1a2867d087a4bfac20a6ca61b82d3 - This change is mandatory, w/o this commit it's not possible to establish Wi-Fi connection during SYS_INIT phase, i.e. when NET_CONFIG library waits for the interface to be up. As most of the upstream samples rely on this feature, we have to support it.
* 2c0359d0fc5ba360a54394543ad535fde46e4cc0 - up for discussion. Since the supplicant selects `NET_SOCKETS_PACKET`, I think it also makes sense to select core `NET_SOCKETS` support, to work out-of-the-box with the sample that do not do it explicitly. I've also noticed that supplicant won't build with the default minimal libc, so change the default to newlib.


krish2718 (approved):

rado17 (approved):

sachinthegreen (approved):

rlubos:
The changes now need to be moved to `sdk-nrf` PR due to upmerge prep.

PR #112, opened 2023-07-17
Fix shell hang when memory is low

krish2718:
When available HEAP is low, the allocation failure to receive the message from control interface can cause the control interface socket to be always "ready" causing the WPA supplicant eloop to a busy loop and no other threads are run (The thread that implements control interface command timeout). If this is from shell, then it results in a shell hang forever.

Use stack to avoid this busy loop.

Fixes WSP-137.

### DIFF
--- a/wpa_supplicant/ctrl_iface_zephyr.c
+++ b/wpa_supplicant/ctrl_iface_zephyr.c
@@ -12,25 +12,21 @@ static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
 					      void *sock_ctx)
 {
 	struct wpa_supplicant *wpa_s = eloop_ctx;
-	char *buf, *pos;
+	char buf[CTRL_IFACE_MAX_LEN + 1], *pos;
 	int res;
 	char *reply = NULL;
 	size_t reply_len = 0;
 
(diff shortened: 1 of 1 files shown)


VivekUppunda (approved):

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -12,25 +12,21 @@ static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
 					      void *sock_ctx)
 {
 	struct wpa_supplicant *wpa_s = eloop_ctx;
-	char *buf, *pos;
+	char buf[CTRL_IFACE_MAX_LEN], *pos;
```
ajayparida, line 15: (1) Should we Initalise here "{0}" instead of doing memset. 
(2) We are allocating 1 byte less, Will it impact anything?

krish2718: 1 - No, for a on-stack buffer both are same (compiler internally does `memset`: see https://stackoverflow.com/a/453462)
2 - done, thanks.

ajayparida (approved):

rado17 (approved):

PR #128, opened 2023-10-19
NCS 2.5.0 rc2 rebase

cvinayak:


krish2718:
Now that we have decided no need for rebasing, can we drop this?

cvinayak:
> Now that we have decided no need for rebasing, can we drop this?

Closing as rebase of hostap is not required for NCS 2.5.0 RC2

PR #130, opened 2023-11-07
[nrf noup] wpa_supplicant: Remove non-output status

kapbh:
There is "OK"/"FAIL" status multipal times during
wifi connect. Skip for wifi management commands.

### DIFF
--- a/wpa_supplicant/wpa_cli_cmds.c
+++ b/wpa_supplicant/wpa_cli_cmds.c
@@ -260,7 +260,7 @@ static int wpa_cli_cmd_disable_network(struct wpa_ctrl *ctrl, int argc,
 static int wpa_cli_cmd_add_network(struct wpa_ctrl *ctrl, int argc,
 				   char *argv[])
 {
-	return wpa_ctrl_command(ctrl, "ADD_NETWORK");
+	return wpa_cli_cmd(ctrl, "ADD_NETWORK", 0, argc, argv);
 }
 
 
@@ -274,19 +274,19 @@ static int wpa_cli_cmd_remove_network(struct wpa_ctrl *ctrl, int argc,
 static int wpa_cli_cmd_disconnect(struct wpa_ctrl *ctrl, int argc,
 				  char *argv[])
 {
-	return wpa_ctrl_command(ctrl, "DISCONNECT");
+	return wpa_cli_cmd(ctrl, "DISCONNECT", 0, argc, argv);
 }
 
 
 static int wpa_cli_cmd_status(struct wpa_ctrl *ctrl, int argc, char *argv[])
 {
 	if (argc > 0 && os_strcmp(argv[0], "verbose") == 0)
-		return wpa_ctrl_command(ctrl, "STATUS-VERBOSE");
+		return wpa_cli_cmd(ctrl, "STATUS-VERBOSE", 0, argc, argv);
 	if (argc > 0 && os_strcmp(argv[0], "wps") == 0)
-		return wpa_ctrl_command(ctrl, "STATUS-WPS");
+		return wpa_cli_cmd(ctrl, "STATUS-WPS", 0, argc, argv);
 	if (argc > 0 && os_strcmp(argv[0], "driver") == 0)
-		return wpa_ctrl_command(ctrl, "STATUS-DRIVER");
-	return wpa_ctrl_command(ctrl, "STATUS");
+		return wpa_cli_cmd(ctrl, "STATUS-DRIVER", 0, argc, argv);
+	return wpa_cli_cmd(ctrl, "STATUS", 0, argc, argv);
 }
 
 
@@ -319,7 +319,7 @@ static int wpa_cli_cmd_interface_add(struct wpa_ctrl *ctrl, int argc,
 	if (os_snprintf_error(sizeof(cmd), res))
 		return -1;
 	cmd[sizeof(cmd) - 1] = '\0';
-	return wpa_ctrl_command(ctrl, cmd);
+	return wpa_cli_cmd(ctrl, cmd, 0, argc, argv);
 }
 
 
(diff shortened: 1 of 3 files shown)


wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -73,16 +74,17 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print,
 		os_memcpy(resp, buf, len - 1);
 	}
 
-	if (print) {
+	if (print_ok) {
 			buf[len] = '\0';
 			printf("%s", buf);
+			print_ok = true;
```
krish2718, line 80: How would this work? As the initial value of `print_ok` is false it would never enter this branch, no?

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -35,13 +35,14 @@ struct wpa_ctrl *mon_conn;
 struct wpa_ctrl *global_ctrl_conn;
 char *ifname_prefix = NULL;
 extern struct wpa_global *global;
+bool print_ok = false;
```
krish2718, line 38: please use `static` and then you can remove the initialisation as well.

wpa_supplicant/wpa_cli_zephyr.c
kapbh: Yes, right ! was missed out.

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -74,17 +75,19 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print,
 		resp[len - 1] = '\0';
 	}
 
-	if (print) {
+	if (!print_ok) {
```
krish2718, line 78: This would print `OK` only once, but what we want is:

* `wpa_cli` should still output OK/FAIL for each command (standard behaviour as this is interactive)
* Everything else i.e., `wifi` commands should never output OK/FAIL, not even once.

krish2718 (changes requested):

sachinthegreen:
Please change this PR to a draft till the final solution is implemented.

wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -260,7 +274,7 @@ static int wpa_cli_cmd_disable_network(struct wpa_ctrl *ctrl, int argc,
 static int wpa_cli_cmd_add_network(struct wpa_ctrl *ctrl, int argc,
 				   char *argv[])
 {
-	return wpa_ctrl_command(ctrl, "ADD_NETWORK");
```
krish2718, line 263: Can you do this in a separate commit as a pre-cursor?

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -78,7 +79,12 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print,
 			buf[len] = '\0';
 			if (buf[0] != '\0')
 				printf("%s", buf);
-	}
+	}else if(!print_ok){
```
krish2718, line 82:
```suggestion
	}else if (!print_ok){
```

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -78,7 +79,12 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print,
...
+			buf[len] = '\0';
+			if (buf[0] != '\0')
+				printf("%s", buf);
+			print_ok = true;
```
krish2718, line 86: Is this needed now?

wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -2632,7 +2646,7 @@ static int wpa_cli_cmd_twt_teardown(struct wpa_ctrl *ctrl, int argc,
 
 static int wpa_cli_cmd_erp_flush(struct wpa_ctrl *ctrl, int argc, char *argv[])
 {
-	return wpa_ctrl_command(ctrl, "ERP_FLUSH");
+	return wpa_ctrl_command_interactive(ctrl, "ERP_FLUSH");
```
krish2718, line 2649: Well, shouldn't these be `wpa_cli_cmd`?

krish2718 (changes requested):

wpa_supplicant/wpa_cli_cmds.c
kapbh: Ok sure.

wpa_supplicant/wpa_cli_zephyr.c
kapbh: This change will print only one time "OK" after flashing the board. Otherwise there is blank line.

jukkar:
Do we really need this? I mean that the OK/FAIL will show the status of the issued command which is kind of useful information.
Annoyingly the FAIL/OK print extra `\n` which looks bad, so I am proposing this change in upstream that skips the extra newline here https://github.com/zephyrproject-rtos/hostap/pull/2/commits/c3dddd36031e838fcae4a31975b96e78393db3a7

krish2718:
> Do we really need this? I mean that the OK/FAIL will show the status of the issued command which is kind of useful information. Annoyingly the FAIL/OK print extra `\n` which looks bad, so I am proposing this change in upstream that skips the extra newline here [zephyrproject-rtos/hostap@c3dddd3](https://github.com/zephyrproject-rtos/hostap/commit/c3dddd36031e838fcae4a31975b96e78393db3a7)

Yes, we had multiple tickets from customers about the random OK/FAIL without any context. These are only useful in interactive environment where a single command is types in and you get a response OK/FAIL with newlines. But in non-interactive environment these are useless and the error checking can be handled programatically.

wpa_supplicant/wpa_cli_zephyr.c
krish2718: We can get rid of this and either print all for interactive or NONE for non-interactive.

kapbh: Okay.

krish2718 (approved):

VivekUppunda (approved):

PR #131, opened 2023-11-08
Fix socket resp handling

krish2718:


kapbh (approved):

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -76,7 +76,8 @@ static int _wpa_ctrl_command(struct wpa_ctrl *ctrl, const char *cmd, int print,
 
 	if (print) {
 			buf[len] = '\0';
-			printf("%s", buf);
+			if (strlen(buf) > 0)
```
jukkar, line 79: Minor nit but we could shorten this to
```
if (buf[0] != '\0')
    printf("%s", buf);
```
This avoids a call to `strlen()`

jukkar (approved):
Just a mental note to the team, we need to remember to port these to upstream when the time comes.

PR #138, opened 2023-12-04
[nrf fromlist] zephyr: Fix possible race when getting scan results

jukkar:
The zephyr driver sets the scan_res2_get_in_prog flag to true after starting to get the results. This can cause possible race as the results might be done before the dev_ops->get_scan_results2() returns. This could lead to timeout in the following while loop.


krish2718 (approved):
Please raise a manifest PR.

jukkar:
Is the manifest changes really needed yet, the original fix is not yet merged and we do not reference that zephyrrtos repo yet in any manifest.

rado17 (approved):

PR #143, opened 2023-12-15
Basic AP mode support

krish2718:
Beaconing only.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -1139,8 +1139,7 @@ static int _wpa_drv_zep_set_key(void *priv,
 		wpa_printf(MSG_ERROR, "%s: Invalid handle\n", __func__);
 		goto out;
 	}
-	if ((alg != WPA_ALG_NONE) &&
-	    ((!seq) || (!key))) {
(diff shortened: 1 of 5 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -1430,6 +1449,42 @@ static int wpa_drv_zep_get_conn_info(void *priv, struct wpa_conn_info *ci)
...
+	if_ctx = priv;
+
+	dev_ops = if_ctx->dev_ctx->config;
+
+	if (!dev_ops->start_ap) {
+		wpa_printf(MSG_ERROR, "%s: set_ap op not supported\n", __func__);
+		goto out;
+	}
+
+	ret = dev_ops->start_ap(if_ctx->dev_priv,
```
sachinthegreen, line 1474: Can we have some comment describing why `set_ap` needs to call `start_ap` ?

sachinthegreen (approved):

src/drivers/driver_zephyr.c
krish2718: Rejigged the ops.

rado17 (approved):

PR #145, opened 2024-01-02
Fix shell freeze

krish2718:


wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -126,8 +126,22 @@ static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
...
+	if (!buf) {
+		/* Expect this to fail in stress test cases, so use MSG_DEBUG
+		 * instead of MSG_ERROR.
+		 */
+		wpa_printf(MSG_DEBUG, "Failed to allocate memory for ctrl_iface receive buffer");
+		/* Drain the socket */
+		res = recv(sock, NULL, 0, MSG_TRUNC);
+		if (res < 0) {
+			wpa_printf(MSG_DEBUG, "recvfrom(ctrl_iface): %s",
+					strerror(errno));
```
sachinthegreen, line 138: Indentation seems to be off.

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -126,8 +126,22 @@ static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
...
+			break;
```
sachinthegreen, line 139: `break;` seems to be invalid here, since it will result in the next `recv` being called with a NULL buf again.

sachinthegreen (changes requested):

sachinthegreen (approved):

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -126,8 +126,21 @@ static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
...
 
 	buf = os_zalloc(CTRL_IFACE_MAX_LEN + 1);
-	if (!buf)
+	if (!buf) {
+		/* Expect this to fail in stress test cases, so use MSG_DEBUG
+		 * instead of MSG_ERROR.
+		 */
+		wpa_printf(MSG_DEBUG, "Failed to allocate memory for ctrl_iface receive buffer");
+		/* Drain the socket */
+		res = recv(sock, NULL, 0, MSG_TRUNC);
```
jukkar, line 135: Is passing `NULL` buf to `recv()` really working as expected, how well is this tested?
I am asking because according to the code, if `buf` is NULL, the packet is requeued back i.e., it is not discarded in this case.
https://github.com/zephyrproject-rtos/zephyr/blob/49f9d8e19c29bd5661bfb5e49c0a4c04871eb054/subsys/net/lib/sockets/sockets.c#L1522
It might be that the poll/select will not trigger again immediately in this case.

krish2718: I had only done basic tests, haven't really tested this scenario as its tough to recreate. I see your point, so, in Zephyr how to discard the socket data? `MSG_TRUNC` shouldn't expect any buffer for len=0, no?

krish2718: > It might be that the poll/select will not trigger again immediately in this case.

That is what is needed.

krish2718: You are right, passing NULL buffer doesn't seem to work, below unit test fails.

```
diff --git a/tests/net/socket/udp/src/main.c b/tests/net/socket/udp/src/main.c
index 690512785eb..9c4e992c196 100644
--- a/tests/net/socket/udp/src/main.c
+++ b/tests/net/socket/udp/src/main.c
@@ -1114,6 +1114,8 @@ void test_msg_trunc(int sock_c, int sock_s, struct sockaddr *addr_c,
                    socklen_t addrlen_s)
 {
        int rv;
+       fd_set rx_set;
+       struct timeval timeout;
        uint8_t str_buf[sizeof(TEST_STR_SMALL) - 1];
 
        rv = bind(sock_s, addr_s, addrlen_s);
@@ -1141,6 +1143,29 @@ void test_msg_trunc(int sock_c, int sock_s, struct sockaddr *addr_c,
        zassert_equal(rv, -1, "consecutive recv should've failed");
        zassert_equal(errno, EAGAIN, "incorrect errno value");
 
+       /* MSG_TRUNC + discard data */
+       rv = send(sock_c, BUF_AND_SIZE(TEST_STR_SMALL), 0);
+       zassert_equal(rv, sizeof(TEST_STR_SMALL) - 1, "send failed");
+
+       rv = recv(sock_s, NULL, 0, ZSOCK_MSG_TRUNC);
+       zassert_equal(rv, 0, "MSG_TRUNC flag failed");
+
+       /* FD should have been cleared */
+       FD_ZERO(&rx_set);
+       FD_SET(sock_s, &rx_set);
+       /* Wait for 1 second */
+       timeout.tv_sec = 1;
+       timeout.tv_usec = 0;
+       rv = select(sock_s + 1, &rx_set, NULL, NULL, &timeout);
+       zassert_equal(rv, 0, "select should've timed out");
+       zassert_false(FD_ISSET(sock_s, &rx_set), "FD_ISSET should've failed");
+
+       /* The entire data should've been discarded */
+       rv = recv(sock_s, str_buf, sizeof(str_buf), ZSOCK_MSG_DONTWAIT);
+       zassert_equal(rv, -1, "consecutive recv should've failed");
+       zassert_equal(errno, EAGAIN, "incorrect errno value");
+
+
        /* MSG_TRUNC & MSG_PEEK combo */
 
        rv = send(sock_c, BUF_AND_SIZE(TEST_STR_SMALL), 0);
```

But NULL buffer with non-zero maximum length worked, select timedout and next `recv` failed.

```
+       /* MSG_TRUNC + discard data + NULL buffer */
+       rv = send(sock_c, BUF_AND_SIZE(TEST_STR_SMALL), 0);
+       zassert_equal(rv, sizeof(TEST_STR_SMALL) - 1, "send failed");
+
+       rv = recv(sock_s, NULL, sizeof(TEST_STR_SMALL), ZSOCK_MSG_TRUNC);
+       zassert_equal(rv, sizeof(TEST_STR_SMALL) - 1, "MSG_TRUNC flag failed");
+
+       /* FD should have been cleared */
+       FD_ZERO(&rx_set);
+       FD_SET(sock_s, &rx_set);
+       /* Wait for 1 second */
+       timeout.tv_sec = 1;
+       timeout.tv_usec = 0;
+       rv = select(sock_s + 1, &rx_set, NULL, NULL, &timeout);
+       zassert_equal(rv, 0, "select should've timed out");
+       zassert_false(FD_ISSET(sock_s, &rx_set), "FD_ISSET should've failed");
+
+       /* The entire data should've been discarded */
+       rv = recv(sock_s, str_buf, sizeof(str_buf), ZSOCK_MSG_DONTWAIT);
+       zassert_equal(rv, -1, "consecutive recv should've failed");
+       zassert_equal(errno, EAGAIN, "incorrect errno value");
```

But from the code the `MSG_TRUNC` flag is only set to indicate the message truncation, but doesn't seem to be used in actual functionality.

rado17 (approved):

wpa_supplicant/ctrl_iface_zephyr.c
krish2718: I will wait for @rlubos to comment before changing the fix to use `NULL buffer + non-zero length`.

rlubos: By no means I'd recommend to provide NULL buf pointer with non-zero length. I hope such case should be caught during arument check, but in no case I think it should cause the data to be flushed on any kind of socket.

First of all, what socket type are we dealing here with? I see you've been checking with UDP test cases, but this file seems to be working with socketpair?

If it's UDP, I'd just provide a dummy one-byte buffer to read - reading a single byte from a datagram socket should discard the rest of the datagram. But if it's socketpair, it seems to be acting as a STREAM socket from what I see. I think the only way to flush the data from the socket would be to call non-blocking `recv()` in a loop (could be with the same single-byte buffer) until it reports EAGAIN.

krish2718: Yes, its socketpair, I used UDP to test `MSG_TRUNC` functionality.

> the only way to flush the data from the socket would be to call non-blocking recv() in a loop

This was my first solution, using a small buffer in `stack` to avoid consuming memory. But later realized that `MSG_TRUNC` might be better. Is there a reason `MSG_TRUNC` isn't implemented for STREAM sockets? Linux `tcp` implements `MSG_TRUN` (See `man 7 tcp`)

rlubos: > Is there a reason MSG_TRUNC isn't implemented for STREAM sockets?

The only reason I can think of is that nobody needed (and thus implemented) it so far. It seems to be Linux-specifc option (i.e. not covered by POSIX man).

krish2718: ok, then I will switch back to draining the socket using smaller buffer on stack, thanks.

wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -126,8 +126,20 @@ static void wpa_supplicant_ctrl_iface_receive(int sock, void *eloop_ctx,
 	size_t reply_len = 0;
 
 	buf = os_zalloc(CTRL_IFACE_MAX_LEN + 1);
-	if (!buf)
+	if (!buf) {
+		/* Do a dummy read to drain the data from the socket */
+		unsigned char dummy[100];
```
jukkar, line 131: You could mark this as static in order to avoid using stack.

jukkar (approved):

PR #146, opened 2024-01-17
Fix SAP restart

krish2718:


src/drivers/driver_zephyr.c
```diff
@@ -1613,6 +1613,7 @@ int wpa_drv_zep_deinit_ap(void *priv)
 	}
 
 out:
+	if_ctx->beacon_set = false;
```
sachinthegreen, line 1616: Should this be protected with `if (if_ctx) {` (since this could crash for the `if (!priv) {` conditional code

src/drivers/driver_zephyr.c
```diff
@@ -1580,8 +1580,8 @@ int wpa_drv_zep_stop_ap(void *priv)
 	}
 
 	if_ctx->freq = 0;
-	if_ctx->beacon_set = false;
 out:
+	if_ctx->beacon_set = false;
```
sachinthegreen, line 1584: Should this be protected with `if (if_ctx) {` (since this could crash for the `if (!priv) {` conditional code

sachinthegreen (changes requested):

rado17 (approved):

sachinthegreen (approved):

PR #149, opened 2024-01-23
[nrf noup] zephyr: Enable AP commands

krish2718:
fixup! [nrf noup] zephyr: Add support for WPA CLI zephyr

AP mode commands were disabled for Zephyr, but now that we support AP mode, enable these commands though not all are used.

Implements SHEL-2411.

### DIFF
--- a/wpa_supplicant/wpa_cli_cmds.c
+++ b/wpa_supplicant/wpa_cli_cmds.c
@@ -351,13 +351,170 @@ static int wpa_cli_cmd_interface_list(struct wpa_ctrl *ctrl, int argc,
 	return wpa_cli_cmd(ctrl, "INTERFACE_LIST", 0, argc, argv);
 }
 
-
-#if !defined(CONFIG_ZEPHYR) || (defined(CONFIG_ZEPHYR) && defined(CONFIG_WPA_CLI))
+#if CONFIG_AP || CONFIG_P2P || !defined(CONFIG_ZEPHYR) || \
+	(defined(CONFIG_ZEPHYR) && defined(CONFIG_WPA_CLI))
 static void wpa_cli_msg_cb(char *msg, size_t len)
 {
 	wpa_printf(MSG_INFO, "%s\n", msg);
 }
+#endif
 
+#ifdef CONFIG_AP
+static int wpa_cli_cmd_sta(struct wpa_ctrl *ctrl, int argc, char *argv[])
+{
+	return wpa_cli_cmd(ctrl, "STA", 1, argc, argv);
+}
+
+
+static char **wpa_cli_complete_sta(const char *str, int pos)
+{
+	int arg = get_cmd_arg_num(str, pos);
+	char **res = NULL;
+
+	switch (arg) {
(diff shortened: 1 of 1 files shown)


wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -3014,6 +3012,31 @@ static const struct wpa_cli_cmd wpa_cli_commands[] = {
 	{ "interface_list", wpa_cli_cmd_interface_list, NULL,
 	  cli_cmd_flag_none,
 	  "= list available interfaces" },
+#ifdef CONFIG_AP
```
jukkar, line 3015: Why do we move this block around, this just makes it harder when working with upstream?

wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -351,13 +351,168 @@ static int wpa_cli_cmd_interface_list(struct wpa_ctrl *ctrl, int argc,
...
 
-
-#if !defined(CONFIG_ZEPHYR) || (defined(CONFIG_ZEPHYR) && defined(CONFIG_WPA_CLI))
+#ifdef CONFIG_AP
 static void wpa_cli_msg_cb(char *msg, size_t len)
 {
 	wpa_printf(MSG_INFO, "%s\n", msg);
 }
 
+static int wpa_cli_cmd_sta(struct wpa_ctrl *ctrl, int argc, char *argv[])
```
jukkar, line 360: Same here, why do we move this code block around?

wpa_supplicant/wpa_cli_cmds.c
krish2718: This was under the `#ifndef !CONFIG_ZEPHYR`, so, I had to move it out of that block.

jukkar: Ok, understood. But if this is to be proposed to upstream and/or when getting a new upstream version, it would be better if we make smaller changes as it helps with conflict resolving. Could we adjust the ifdef statements so that we do not need to move the code?

krish2718: Upstream doesn't even have this file, I have carve it out of `wpa_cli.c` to avoid duplication.

> Could we adjust the ifdef statements so that we do not need to move the code?

It's tricky, as its in the middle as below.
```
#ifndef !CONFIG_ZEPHYR
..
..
...
#ifdef CONFIG_AP
...
...
#endif
....
...
#endif
```

sachinthegreen (approved):

rado17 (approved):

wpa_supplicant/wpa_cli_cmds.c
jukkar: > Upstream doesn't even have this file, I have carve it out of `wpa_cli.c` to avoid duplication.

Ok, I thought it had because I did not see "zephyr" mentioned in the file name. No worries then, I was mostly concerned about the future merges.

jukkar (approved):

PR #150, opened 2024-01-25
Control interface improvements

krish2718:
To handle OOM conditions in a SnS test (Stress and Stability) after few hours.

### DIFF
--- a/wpa_supplicant/ctrl_iface.c
+++ b/wpa_supplicant/ctrl_iface.c
@@ -13003,7 +13003,7 @@ static int wpas_global_ctrl_iface_fst_detach(struct wpa_global *global,
 char * wpa_supplicant_global_ctrl_iface_process(struct wpa_global *global,
 						char *buf, size_t *resp_len)
 {
-	static char reply[2048];
+	static char reply[1024];
 	const int reply_size = sizeof(reply);
 	int reply_len;
 	int level = MSG_DEBUG;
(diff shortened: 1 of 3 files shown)


wpa_supplicant/ctrl_iface_zephyr.c
```diff
@@ -277,18 +277,37 @@ static void wpa_supplicant_global_ctrl_iface_receive(int sock, void *eloop_ctx,
 	size_t reply_len = 0;
 
 	buf = os_zalloc(CTRL_IFACE_MAX_LEN + 1);
-	if (!buf)
+	if (!buf) {
+		/* Do a dummy read to drain the data from the socket */
+		static unsigned char dummy[512];
```
jukkar, line 282: As the temp buffer is just meant for draining the socket, we can use much smaller buffer here (it will just loop more during read), and then we do not need to make this static (saves some memory). I suggest something like 32 or 64 bytes long buffer here.

krish2718: Sure, we can do that. But next set of changes is to move from heap->stack, hence made it draft, that makes all these hacks moot. But would require changing a lot of open-source code, so, still thinking on it.

sachinthegreen (approved):

wpa_supplicant/ctrl_iface.c
```diff
@@ -13009,10 +13003,13 @@ static int wpas_global_ctrl_iface_fst_detach(struct wpa_global *global,
 char * wpa_supplicant_global_ctrl_iface_process(struct wpa_global *global,
 						char *buf, size_t *resp_len)
 {
-	char *reply;
-	const int reply_size = 2048;
+	static char reply[1024];
+	const int reply_size = 1024;
```
jukkar, line 13007: Better
```
const int reply_size = sizeof(reply);
```
So that `reply_size` will always be correct and we do not need to sync the value if `reply` size changes.

jukkar (approved):

PR #151, opened 2024-01-29
[nrf noup] Update device name to nrf_wlan0

rado17:
Read device name from supplicant context instead of using hard-coded strings

### DIFF
--- a/wpa_supplicant/wpa_cli_zephyr.c
+++ b/wpa_supplicant/wpa_cli_zephyr.c
@@ -113,7 +113,7 @@ static void wpa_cli_close_connection(struct wpa_supplicant *wpa_s)
 	wpa_s->ctrl_iface->mon_sock_pair[1] = -1;
 }
 
-static void wpa_cli_recv_pending(struct wpa_ctrl *ctrl)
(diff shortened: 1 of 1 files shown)


jukkar (changes requested):
Hard coding something like the network interface name should be avoided. Better would be to get the network interface name from the used network interface and use that. Then when the device boots, the interface name could be set to `nrf_wlan0`.
You are now basically forking the code as this cannot be accepted upstream, and it will conflict the in future upstreaming efforts.

wpa_supplicant/wpa_cli_zephyr.c
```diff
@@ -27,7 +27,7 @@
 #define CMD_BUF_LEN  1024
 #define MAX_CMD_SIZE 512
 #define MAX_RESPONSE_SIZE 512
-#define DEFAULT_IFNAME "wlan0"
+#define DEFAULT_IFNAME "nrf_wlan0"
```
jukkar, line 30: Please remove this setting, the default name should be ok here.

rado17: Done.

jukkar (approved):

sachinthegreen (approved):

PR #155, opened 2024-02-25
Fix/assoc resp ies

krish2718:


sachinthegreen (approved):

VivekUppunda:
Just a generic comment. Would it be better to move that function - wpa_supplicant_event_wrapper to a switch statement block at some future time?

VivekUppunda (approved):

krish2718:
> Just a generic comment. Would it be better to move that function - wpa_supplicant_event_wrapper to a switch statement block at some future time?

Sure, but compiler will do it anyway. I have raised a Jira to revamp this entire design, we can do it then. Thanks.

VivekUppunda:
> > Just a generic comment. Would it be better to move that function - wpa_supplicant_event_wrapper to a switch statement block at some future time?
> 
> Sure, but compiler will do it anyway. I have raised a Jira to revamp this entire design, we can do it then. Thanks.

Hi Chaitanya, 
As far as I have read previously and understand, a switch table is normally created into a jump table by most compilers while an if-else statement would normally be created into a binary tree. What I have read is that if the number of elements in the check is less - an if-else statement is faster but if the number of elements to check progressively increases a switch statement is faster. Whether speed is a requirement here, i think we can always decide if it matters in this case.  
But, in this case the compiler is optimizing it, i think it is very good

Thanks
Vivek

PR #158, opened 2024-03-04
Fixes for upmerge 01.03.2024

rlubos:
~~For now fixing twister only, it needs to be verified with downstream tests whether the change actually makes sense.~~
Good to go from my PoV.

### DIFF
--- a/src/utils/includes.h
+++ b/src/utils/includes.h
@@ -49,11 +49,13 @@
 #include <zephyr/posix/sys/select.h>
 #include <zephyr/posix/sys/socket.h>
 #include <zephyr/posix/unistd.h>
+#include <zephyr/posix/signal.h>
 #else /* defined(CONFIG_POSIX_API) */
 #include <zephyr/net/net_ip.h>
 #include <zephyr/net/socket.h>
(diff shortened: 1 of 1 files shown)


jukkar (approved):

src/utils/includes.h
```diff
@@ -49,11 +49,13 @@
...
 #include <zephyr/posix/unistd.h>
+#include <zephyr/posix/signal.h>
 #else /* defined(CONFIG_POSIX_API) */
 #include <zephyr/net/net_ip.h>
 #include <zephyr/net/socket.h>
 #endif /* defined(CONFIG_POSIX_API) */
 #include <zephyr/shell/shell.h>
 #endif /* defined(CONFIG_ZEPHYR) */
 
+#define signal(a, b) (void)(b)
```
krish2718, line 60: should this be for Zephyr only? or at least `#ifndef signal`

rlubos: I'll add ifndef

sachinthegreen (approved):

src/utils/includes.h
rlubos: done

PR #160, opened 2024-03-06
Log errors to check 4-way handshake failure

krish2718:


jukkar (approved):

SeppoTakalo:
Is this relevant for the release?
Seems quite minor change for just the logging.

krish2718:
> Is this relevant for the release? Seems quite minor change for just the logging.

Its minor, but this is based on feedback we have got from customers, and was missed in RC1, having errors logs help us quickly identify the issue as debug logs are disabled due to flash constraints.

krish2718:
@SeppoTakalo  please confirm if I can go ahead and merge it?

SeppoTakalo:
Why is there no upstream PR?

I don't understand how the hostapp development model works, but my first assumption is that this is not a change that justifies `[nrf noup]` changes to be cumulated in our Git history.

krish2718:
> Why is there no upstream PR?
> 
> I don't understand how the hostapp development model works, but my first assumption is that this is not a change that justifies `[nrf noup]` changes to be cumulated in our Git history.

You are right, we do follow upstream first approach. But this was a bit last moment, I can convert to `to up` (though not supposed be used often) or we can drop this from 2.6.0, I will leave it to you.

SeppoTakalo:
Drop from 2.6.0 as we are just trying to create RC.
Too many manifest changes still coming in.

Send this to upstream.

PR #161, opened 2024-03-07
[nrf noup] Conditionally compile wpa commands help strings.

markaj-nordic:
This saves FLASH memory in cases when the wpa cli is not really used.

This patch essentially copies the `wpa_cli_cmd` array without help strings and pulls it
to the build process instead of the already existing set of "full" commands when the `CONFIG_WPA_CLI`
is not selected.

As a followup this could be refactored, e.g. help strings could be stored as an independent object, that could be
easily overwritten with NULLs (wpa_cli_cmd would then reference the strings by IDs etc.).

In the case of Matter if saves 0.7KB (enough to unblock nightly CI). For applications that utilize advanced supplicant features it's much more.


wpa_supplicant/wpa_cli_cmds.c
```diff
@@ -3646,6 +3647,373 @@ static const struct wpa_cli_cmd wpa_cli_commands[] = {
 #endif  /* !CONFIG_ZEPHYR || (CONFIG_ZEPHYR && CONFIG_WPA_CLI)*/
 	{ NULL, NULL, NULL, cli_cmd_flag_none, NULL }
 };
+#else
+/* Compile commands without help string to spare some bytes */
+static const struct wpa_cli_cmd wpa_cli_commands[] = {
+	{"status", wpa_cli_cmd_status, NULL, cli_cmd_flag_none, NULL},
```
LuDuda, line 3653: Not sure what will be easier to maintain in future.. but alternatively you can do:

```
#ifdef CONFIG_WPA_CLI
#define HELP_MSG(msg) (msg)
#else
#define HELP_MSG(msg) NULL
#endif
```
And later:
```
{ "status", wpa_cli_cmd_status, NULL,
	  cli_cmd_flag_none,
	  HELP_MSG("[verbose] = get current WPA/EAPOL/EAP status") },
```

LuDuda (commented):
Thanks :+1:

VivekUppunda (approved):

rado17 (approved):

jfischer-no:
Really 700bytes of flash? How long will it take, one, two releases? Next matter revision update? Anyway, too fragile for me.

markaj-nordic:
> Really 700bytes of flash? How long will it take, one, two releases? Next matter revision update? Anyway, too fragile for me.

Currently the Windows CI builds fail because of 20 bytes missing. And 700 is for Matter which uses only short set of wpa commands. I guess for more WiFi advanced apps, it's much more, yet it's only stripping off strings that are not used anyway. 

Don't worry, we will enable LTO for this application post 2.6.0. It will save ~50-60 KB.

wpa_supplicant/wpa_cli_cmds.c
krish2718: Yeah, using a macro to compile out RO strings is good. And this commands is `nrf noup` anyway, as I didn't want to duplicate the commands, so, moved to a new file `wpa_cli_cmds.c` which doesn't exist in the opensource, so, this is fine.

PerMac:
@markaj-nordic 
> Don't worry, we will enable LTO for this application post 2.6.0. It will save ~50-60 KB.

Did that happen? Is this PR still needed?

krish2718:
@markaj-nordic is this PR still needed? I guess not, so, can you please close?

markaj-nordic:
> @markaj-nordic is this PR still needed? I guess not, so, can you please close?

Closed, but still I think I might be good to revisit this, especially in the context of samples that uses advanced features of the supplicant.

PR #162, opened 2024-03-19
[nrf noup] wpa_supplicant: Add connection event for wrong key

kapbh:
When there is a wrong password for AP, the send_wifi_mgmt_conn_event is missing for WRONG_KEY.

### DIFF
--- a/wpa_supplicant/events.c
+++ b/wpa_supplicant/events.c
@@ -3674,6 +3674,9 @@ static void wpa_supplicant_event_disassoc_finish(struct wpa_supplicant *wpa_s,
 		if (wpas_p2p_4way_hs_failed(wpa_s) > 0)
 			return; /* P2P group removed */
 		wpas_auth_failed(wpa_s, "WRONG_KEY");
+#ifdef CONFIG_ZEPHYR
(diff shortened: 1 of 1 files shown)


wpa_supplicant/events.c
```diff
@@ -3674,6 +3674,9 @@ static void wpa_supplicant_event_disassoc_finish(struct wpa_supplicant *wpa_s,
 		if (wpas_p2p_4way_hs_failed(wpa_s) > 0)
 			return; /* P2P group removed */
 		wpas_auth_failed(wpa_s, "WRONG_KEY");
+#ifdef CONFIG_ZEPHYR
+		send_wifi_mgmt_conn_event(wpa_s, reason_code);
```
krish2718, line 3678: Reason code here doesn't necessarily translate to WRONG_KEY, so, may be just send `WLAN_REASON_4WAY_HANDSHAKE_TIMEOUT`

kapbh: Okay.

krish2718 (approved):

jukkar (approved):

PR #164, opened 2024-03-22
[nrf noup] zephyr: Start DHCP client only in station mode

rado17:
DHCP client is needed only when the device is running in station mode. Disable this when device is running as AP.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -1525,10 +1525,14 @@ static int wpa_drv_zep_set_supp_port(void *priv,
 				     authorized,
 				     if_ctx->bssid);
 
-#ifdef CONFIG_NET_DHCPV4
-	if (authorized) {
-		net_dhcpv4_restart(iface);
-    }
+#if defined(CONFIG_NET_DHCPV4) && defined(CONFIG_SME)
+	struct wpa_supplicant *wpa_s = if_ctx->supp_if_ctx;
+	/* Need DHCP client in STA mode only */
+	if (wpa_s && wpa_s->sme.ssid_len > 0) {
+		if (authorized) {
+			net_dhcpv4_restart(iface);
+		}
+	}
 #endif
 
 	return ret;


src/drivers/driver_zephyr.c
```diff
@@ -1507,7 +1507,7 @@ static int wpa_drv_zep_set_supp_port(void *priv,
 				     authorized,
 				     if_ctx->bssid);
 
-#ifdef CONFIG_NET_DHCPV4
+#if !defined(CONFIG_AP) && defined(CONFIG_NET_DHCPV4)
```
krish2718, line 1510: This should be `CONFIG_STA` mode (need to introduce it), as we might have other cases like P2P without AP.

rado17: I see that CONFIG_P2P selects CONFIG_AP. So, this check should also cover P2P case as well, right?

sachinthegreen (approved):

src/drivers/driver_zephyr.c
```diff
@@ -1525,10 +1525,14 @@ static int wpa_drv_zep_set_supp_port(void *priv,
 				     authorized,
 				     if_ctx->bssid);
 
-#ifdef CONFIG_NET_DHCPV4
-	if (authorized) {
-		net_dhcpv4_restart(iface);
-    }
+#if defined(CONFIG_NET_DHCPV4) && defined(CONFIG_SME)
```
krish2718, line 1528: SME is mainly for separate auth and assoc compared to a single connect. And also as I said in https://github.com/nrfconnect/sdk-nrf/pull/14718#discussion_r1617261406 this needs to be per-iface setting which is a bit complicated.

Thinking on this, this should really be application's responsibility, I added this in WPA supp as a hack to be common to all sample. WPA supplicant shouldn't be managing DHCP at all. May be just remove?

@rlubos the main reason for doing this was DHCP starts right after iface up, and by the time Wi-Fi is connected, it has hit 1min random backoff, so, any attempt over Wi-Fi reconnection takes about 1min.

DHCP client should ideally register for NET_IF events and always (re)start on UP_RUNNING, WDYT?

krish2718: I would recommend to add `CONFIG_NET_DHCPV4=n` in SAP configuration for time being and we can discuss on a proper solution that is part of the networking stack/application.

krish2718 (changes requested):

src/drivers/driver_zephyr.c
rlubos: > DHCP client should ideally register for NET_IF events and always (re)start on UP_RUNNING, WDYT?

It actually does that already (https://github.com/zephyrproject-rtos/zephyr/blob/53ddff639562ef68dc0a6f62b82f7505c75ebdce/subsys/net/lib/dhcpv4/dhcpv4.c#L1417), however the dhcpv4 have to be started already on the interface anyway.

krish2718: Thanks @rlubos , should we add a Kconfig option to do this on boot for the default iface, once it's UP_RUNNING? This should server for most usecases.

@rado17 We should test and explore this by removing the DHCP restart from WPA supplicant?

rlubos: Couldn't that be achieved by enabling`NET_CONFIG_SETTINGS`, but w/o any other configurations (I. e. no static address etc.). It should then start the DHCP client only. @jukkar thoughts?

jukkar: > Couldn't that be achieved by enabling`NET_CONFIG_SETTINGS`, but w/o any other configurations (I. e. no static address etc.). It should then start the DHCP client only

Yes, that is one possibility, but note that config lib is optional and if it is not enabled, application needs to start dhcp client manually.

krish2718: As removing this from WPA supplicant is invasive, as samples today rely on WPA supplicant to manage DHCP client, let's go ahead with runtime SME check and then in next release we can remove this and let samples manage DHCP.

src/drivers/driver_zephyr.c
rado17: Removed the config dependencies that were added here.

wpa_supplicant/Makefile
```diff
@@ -403,6 +403,10 @@ OBJS += wifi_display.o
 endif
 endif
 
+ifdef CONFIG_STA
```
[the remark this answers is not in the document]
rado17: Yes. Missed removing this when approach was changed. Removed it now.

krish2718 (approved):

PR #165, opened 2024-04-01
[nrf noup] zephyr: Improve error logging

krish2718:
fixup! [nrf noup] zephyr: Redesign the events passing to wpa_supplicant from driver

Useful in debugging.

### DIFF
--- a/src/drivers/driver_zephyr.c
+++ b/src/drivers/driver_zephyr.c
@@ -50,7 +50,9 @@ void wpa_supplicant_event_wrapper(void *ctx,
 				char *ies = os_zalloc(data->auth.ies_len);
 
 				if (!ies) {
-					wpa_printf(MSG_ERROR, "%s: Failed to alloc ies\n", __func__);
+					wpa_printf(MSG_ERROR,
+					  "%s:%d Failed to alloc %d bytes\n",
+					  __func__, __LINE__, data->auth.ies_len);
 					return;
 				}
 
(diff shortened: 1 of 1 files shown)


src/drivers/driver_zephyr.c
```diff
@@ -50,7 +50,9 @@ void wpa_supplicant_event_wrapper(void *ctx,
 				char *ies = os_zalloc(data->auth.ies_len);
 
 				if (!ies) {
-					wpa_printf(MSG_ERROR, "%s: Failed to alloc ies\n", __func__);
+					wpa_printf(MSG_ERROR,
+					  "%s:%d Failed to alloc ies: (%d)bytes\n",
```
jukkar, line 54: Small nit. The output string looks a bit hard to read, why not have `...: %d bytes\n` instead?

src/drivers/driver_zephyr.c
```diff
@@ -50,7 +50,9 @@ void wpa_supplicant_event_wrapper(void *ctx,
 				char *ies = os_zalloc(data->auth.ies_len);
 
 				if (!ies) {
-					wpa_printf(MSG_ERROR, "%s: Failed to alloc ies\n", __func__);
+					wpa_printf(MSG_ERROR,
+					  "%s:%d Failed to alloc ies: %d bytes\n",
```
jukkar, line 54: Another small nit. If we want to save flash, we could construct the string so that only one copy is used for all these printouts, like this:
```
wpa_printf(MSG_ERROR,
	   "%s:%d Failed to alloc %s: %d bytes\n",
	   __func__, __LINE__, "ies", data->auth.ies_len);
```

krish2718: We can even drop the `prefix` as we have line number.

jukkar (approved):

sachinthegreen (approved):

PR #172, opened 2024-05-28
[nrf toup] Zephyr: Fix the data type for time

krish2718:
We use subtraction operation over time, so, it has to be signed.

Fixes "[nrf toup] Use 64bit data type to store time".

Fixes SHEL-2731.

### DIFF
--- a/src/utils/os.h
+++ b/src/utils/os.h
@@ -9,7 +9,7 @@
 #ifndef OS_H
 #define OS_H
 
-typedef uint64_t os_time_t;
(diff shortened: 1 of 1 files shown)


sachinthegreen (approved):

jukkar (approved):

krish2718:
Working with @carlescufi  to fix the CI to accept `nrf toup`.

krish2718:
Overriding compliance as it's only about missing sauce tag `nrf toup`.

PR #178, opened 2024-09-26
Psa prf 256 bit implementation

VivekUppunda:
This PR brings in PRF-256 implementation using PSA APIs for 11.6.1.7.2 in 802.11ac to crypto_mbedtls_alt.c. The PSA wrapper API has been verified via Test vectors derived during connection and validated against multiple derived test vectors


VivekUppunda:
I unintentionally opened in the wrong repo looks like. closing the commit

